PR #6, opened 2021-01-01
Fixed delay to use AHB clock / 8 instead of CORE clk

Geens:
I noticed delays where off by a factor 8. This was because the delay function calculated it's time based off the core clock. The cortex system timer is actually connected to AHB clock with a prescaler of /8.


summner:
Confirmed working - also see: https://stackoverflow.com/questions/64172446/systick-timer-on-cortex-m4-what-is-its-prescaler
@dotcypress Can you take a look ?

dotcypress:
Awesome, thanks!

PR #8, opened 2021-01-30
Use Delay implementation from cortex-m crate.

pawelchcki:
While working on other changes I noticed the latest `cortex-m` crate  got the Delay implementation that is tiny bit more  powerful than the current one in this crate. So I updated the dependency and removed most of Delay implementation, I left the `time` based function as a trait for the main `Delay`, and added them to `prelude`, not sure if that is the best approach so any comments highly welcome. 

The implementation from `cortex-m `crate uses the Core-provided clock
for the timer instead of the default External this gives 8 times better resolution.

Additionally the crate's implementation can handle much longer delays
as it uses 64 bit integer to calculate needed cycles for sleep.

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -1,91 +1,29 @@
-//! Delays
-use core::cmp;
-use cortex_m::peripheral::SYST;
-use hal::blocking::delay::{DelayMs, DelayUs};
-
-use crate::prelude::*;
 use crate::rcc::Clocks;
-use crate::time::{Hertz, MicroSecond};
+use crate::time::MicroSecond;
+pub use cortex_m::delay::*;
+use cortex_m::{peripheral::SYST, prelude::_embedded_hal_blocking_delay_DelayUs};
 
-/// System timer (SysTick) as a delay provider
-pub struct Delay {
-    clk: Hertz,
-    syst: SYST,
+pub trait SYSTDelayExt {
+    fn delay(self, clocks: &Clocks) -> Delay;
 }
 
-impl Delay {
-    /// Configures the system timer (SysTick) as a delay provider
-    pub fn new(syst: SYST, clocks: &Clocks) -> Self {
-        Delay {
-            syst,
-            clk: clocks.ahb_clk / 8,
-        }
-    }
-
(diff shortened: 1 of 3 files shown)


src/delay.rs
```diff
@@ -1,91 +1,29 @@
...
-    fn delay_ms(&mut self, ms: u32) {
-        self.delay_us(ms.saturating_mul(1_000));
-    }
-}
-
-impl DelayMs<u16> for Delay {
-    fn delay_ms(&mut self, ms: u16) {
-        self.delay_ms(ms as u32);
-    }
+pub trait SYSTDelayExt {
```
pawelchcki, line 6: Renamed this because the `DelayExt` would fit more for extensions of `Delay` since this crate isn't yet widely used and is `v0.0.0`, changes like this could be acceptable.

src/delay.rs
```diff
@@ -1,91 +1,29 @@
...
+    fn delay<T>(&mut self, delay: T)
+    where
+        T: Into<MicroSecond>;
 }
 
-impl DelayExt for SYST {
-    fn delay(self, clocks: &Clocks) -> Delay {
-        Delay::new(self, clocks)
+impl DelayExt for Delay {
+    fn delay<T>(&mut self, delay: T)
```
pawelchcki, line 23: I contemplated removing this function, but `time` is a nicer abstraction than simple `u32`. Possibly in the future if `cortex-m` starts using `embedded-time` they will have this interface too.

pawelchcki:
@dotcypress, I switched the Delay implementation to use one from cortex_m. Its not a critical change :) but brings in some nice improvements over the previous implementation - namely around handling larger intervals. Can you take look ?

dotcypress:
@pawelchcki yup, will take a look but a bit later today or tomorrow.
Anyway thank you for helping!

pawelchcki:
Thanks @dotcypress! Absolutely no rush on this PR as its mostly cosmetics. But figured I'll use my weekly quota of pinging You directly on it :joy:. 

>thank you for helping!

Likewise!

pawelchcki:
Thanks !

PR #11, opened 2021-04-11
Gpio alternate functions as typestate

### LINKED ISSUE #10: GPIO alternate function not configurable
It's  not possible to configure the alternate function registers of the GPIO pins. This is needed for implementing peripherals like I2C and others.

The STM32F3 series has the same register interface for GPIO. I'll try and port their [implementation](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/gpio.rs) and [code generator ](https://github.com/stm32-rs/stm32f3xx-hal/tree/master/codegen). I'm making this issue for tracking and discussion.

Geens:
Closes #10


Geens:
I was working with this branch trying to get I2C to work and noticed the output type (push-pull / open-drain) can be controlled separately from the alternate function. This means the typestates for alternate function and other pin configurations should not be combined but should be separate. I'll fix this, so please hold off on merging until I've had the time to do so.

luctius:
@Geens : Are you still working on this; I have my own modifications to gpio, and an adc driver which depends on those changes, I would like to submit, but would like to avoid a major merge conflict. If however, you are working on it for the foreseeable future I will make my own pull request shortly.

Geens:
Hi @luctius, I don't think I will finish this any time soon. I'll close the pull request for now and start over if I come back to it.

PR #13, opened 2021-07-19
[2] Sysconfig

luctius:
Adds basic Sysconfig Support

### DIFF
--- a/src/bb.rs
+++ b/src/bb.rs
@@ -1,0 +1,48 @@
+//! Bit banding
+//!
+//! Support for the manipulation of peripheral registers through bit-banding.
+//! Not all peripherals are mapped to the bit-banding alias region, the peripheral bit-band region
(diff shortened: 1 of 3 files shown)


luctius:
@dotcypress, Could any of you take a look at the other pull requests too please?

dotcypress:
@luctius wow! There are a lot of PR's, thank you!
I can take a look later today.

I'm not have a lot of free time for stm32g4xx and I'm thinking about adding you as collaborator here, what do you think?

luctius:
> @luctius wow! There are a lot of PR's, thank you!
> I'm not have a lot of free time for stm32g4xx and I'm thinking about adding you as collaborator here, what do you think?

We are a small team working on stm32g4xxx and have more work incoming (SPI and FdCAN being directly in the pipeline). My hope is to generate enough momentum that other people start using this and we can in turn benefit from their improvements, so I would be more than willing to spend the time in this repository.

dotcypress:
Great! 🦀

Invite sent.

PR #26, opened 2021-08-26
Support for I2C

mgottschlag:
This PR adds support for I2C. Most of the code is taken from `stm32g0xx-hal`, with only minor changes:
- Clock gating/reset works differently.
- Pins are specified in a different way.
- Documentation is improved.

I personally do not like the way the bus is configured - `with_timing()` imho is not the kind of abstraction that should be in a user-facing API. I guess that there are workflows that require this API, though.

The code is not yet tested - it should be ready for merging once I have performed some basic tests.


mgottschlag:
After I fixed a bug in the original code (see https://github.com/stm32-rs/stm32g0xx-hal/pull/69), the pull request should now be ready for review/merging. For testing, I executed the i2c-mpu6050 example and verified with an oscilloscope that the bit rate is correct.

mgottschlag:
Oh, and tell me if you want the commits to be squashed, as I usually do.

luctius:
This looks good to me, thank you for your effort!

Squashing can be done from by Github.

PR #29, opened 2021-08-29
Support for PWM

mgottschlag:
This PR adds support for PWM. I took the code from stm32h7xx-hal as the basis as it seemed to provide the most complete implementation for the various timers. The changes include:
- stm32g4xx-hal uses a different API for clock gating/reset.
- stm32g4 is missing many bit/field names, so the code often uses .bits()
  or .set_bit()/.clear_bit() instead.
- The advanced-control timers of the STM32G4 MCUs provide four
  complementary outputs.

**This PR requires a recent git version of stm32g4 as the last released version contains some wrong register sizes and types.** Therefore, this PR cannot be merged as-is.

I tested a single PWM as shown in the example. The oscilloscope showed the correct duty cycle and period. I did not test advanced features such as deadtime insertion or complementary outputs. I will conduct a test with multiple PWM outputs later.


luctius:
A new version of the stm32g4 crate has been merged, does that resolve your problems regarding this PR?

mgottschlag:
Yes, it does. This is probably ready to be merged.

PR #30, opened 2021-08-30
Draft: Initial FDCAN support

luctius:
Initial support for CAN and FDCAN.

The CAN and FDCAN are two modules, and I'm still unsure how to do this. Part of it is taken from BxCAN, and the idea is that with a bit of effort FDCAN could be a separate crate as well, but with how the Message RAM works this seems to be more difficult than with BxCAN.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ cortex-m = "0.7.1"
 nb = "0.1.1"
 stm32g4 = "0.13.0"
 paste = "1.0"
+bitflags = "1.2"
+vcell = "0.1"
+static_assertions = "1.1"
 
 [dependencies.cast]
 version = "0.2.7"
@@ -40,6 +43,10 @@ version = "1.0.2"
 default-features = false
 version = "1.1"
 
+[dependencies.defmt]
+version = "0.2.3"
+optional = true
+
 [dev-dependencies]
 cortex-m-rt = "0.6.10"
 cortex-m-rtfm = "0.5.1"
@@ -55,6 +62,7 @@ rtt-target = { version = "0.3.0", features = ["cortex-m"] }
 panic-rtt-target = { version = "0.1.1", features = ["cortex-m"] }
 mpu6050 = "0.1.4"
 
+#TODO: Separate feature sets
 [features]
 default = ["rt"]
 rt = ["stm32g4/rt"]
@@ -70,6 +78,7 @@ stm32g4a1 = ["stm32g4/stm32g4a1"]
 log-itm = ["cortex-m-log/itm"]
 log-rtt = []
 log-semihost = ["cortex-m-log/semihosting"]
+unstable-defmt = ["defmt"]
 
 [profile.dev]
 codegen-units = 1
--- a/examples/can-echo.rs
+++ b/examples/can-echo.rs
@@ -1,0 +1,150 @@
+#![no_main]
+#![no_std]
+
+use crate::hal::{
+    fdcan::{
+        config::NominalBitTiming,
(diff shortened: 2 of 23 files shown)


mgottschlag:
I've written a tool to print the pin mappings for user-specified peripherals, as I did not find anything similar before. This tool results in the following output for CAN:

```
FDCAN1_RX:
  PA11 (STM32G431, STM32G441, STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
  PB8-BOOT0 (STM32G431, STM32G441, STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
  PD0 (STM32G431, STM32G441, STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
FDCAN1_TX:
  PA12 (STM32G431, STM32G441, STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
  PB9 (STM32G431, STM32G441, STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
  PD1 (STM32G431, STM32G441, STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
FDCAN2_RX:
  PB12 (STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
  PB5 (STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
FDCAN2_TX:
  PB13 (STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
  PB6 (STM32G471, STM32G473, STM32G474, STM32G483, STM32G484, STM32G491, STM32G4A1)
FDCAN3_RX:
  PA8 (STM32G473, STM32G474, STM32G483, STM32G484)
  PB3 (STM32G473, STM32G474, STM32G483, STM32G484)
FDCAN3_TX:
  PA15 (STM32G473, STM32G474, STM32G483, STM32G484)
  PB4 (STM32G473, STM32G474, STM32G483, STM32G484)
```

If this is helpful to you, see (https://github.com/mgottschlag/stm32-pin-tools).

Is CAN limited to these specific pin pairs, or is it possible to mix any RX and TX pin, just like with other peripherals?

luctius:
Oh, that is helpful indeed, thanks!

> Is CAN limited to these specific pin pairs, or is it possible to mix any RX and TX pin, just like with other peripherals?

Good point, I need to implement it more generally, though my first focus is to get this actually working ;-).

luctius:
FDCAN is working now; and I've reorganised things a bit.

Since the PR is rather big, some comments would be welcome.

I do think that some things can be simplified somewhat, but overall I think it is a good starting point.

richardeoin:
I copied this FDCAN support to use on the STM32H7, and with a few minor tweaks I was able to get the basic echo example working on a STM32H735. [PR here](https://github.com/stm32-rs/stm32h7xx-hal/pull/274). As a courtesy I also made an issue on this repo https://github.com/stm32-rs/stm32g4xx-hal/issues/34 with some items I noticed whilst reading the code.

My preference would be to move this FDCAN support into a common crate at [stm32-rs/fdcan](https://github.com/stm32-rs/fdcan) so that bugfixes and developments can be used by all devices. If no one has any objections, I will create a skeleton project there with the intention of copying this FDCAN support over. Collaboration would be very welcome!

PR #35, opened 2021-11-11
Added the alternate push pull function

NandoBongers:


mgottschlag:
I believe this is wrong - the function does what the existing function `into_alternate()` is supposed to do. However, maybe I am overlooking something.

In any case, the modification to `OTYPER` has to be present in `into_alternate()` as well, because that function must not produce a open-drain pin if the same pin was previously used with `into_alternate_open_drain()`.

If I am overlooking something, please explain what you had in mind with this PR. :)

NandoBongers:
Hello @mgottschlag I renamed the functions as you suggested in my latest commit.

luctius:
I see no further issues with this commit.

Issue #34, opened 2021-11-06
FDCAN issues


richardeoin:
I happened to spot some (potential) issues in the FDCAN support

* `TXBCR` `TXBCF` and `TXBTO` registers are bitmaps, not ordinals [fdcan.rs#L1396](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan.rs#L1396)
* `TXBRP` register is a bitmap [fdcan.rs#L1415](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan.rs#L1415)
* Tx buffer element `ID_W` is missing its top bit [txbuffer_element.rs#L130](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan/message_ram/txbuffer_element.rs#L130) See RM0440 rev5 44.3.6
* `accept_all_into_fifo1()` has the action for FIFO0 [filter.rs#L40](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan/filter.rs#L40) [filter.rs#L83](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan/filter.rs#L83)

David-OConnor:
Related: Thoughts on a standalone fdCAN crate, like the bxCAN one?

teamplayer3:
Implementation suggestions: Would be nice to get the raw ID from `ExtendedId` and `StandardId` from outside the crate. Maybe make `as_raw()` public or some other way. 
A nice way would be to have a crate to abstract this for every can implementation, so the ID types can be used in crates which should support any can implementation.

luctius:
> I happened to spot some (potential) issues in the FDCAN support
> 
>     * `TXBCR` `TXBCF` and `TXBTO` registers are bitmaps, not ordinals [fdcan.rs#L1396](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan.rs#L1396)
> 
>     * `TXBRP` register is a bitmap [fdcan.rs#L1415](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan.rs#L1415)
> 
>     * Tx buffer element `ID_W` is missing its top bit [txbuffer_element.rs#L130](https://github.com/stm32-rs/stm32g4xx-hal/blob/master/src/fdcan/message_ram/txbuffer_element.rs#L130) See RM0440 rev5 44.3.6
> 
[...]

PR #39, opened 2021-12-13
Transition to the fdcan-rs crate

luctius:
As discussed in #34, the FDCAN part has been transitioned to a new crate. This PR uses that crate instead of our own implementation.


kevswims:
@luctius is there anything I can do to help get this in? I have a board with an STM32g4 and 2 CAN transceivers on it that I can do testing on if needed. At the moment I am using a super hacky driver in my own code to get the CAN working but I would love to move to this if possible.

luctius:
I've updated this to fdcan 0.1.2. It builds; but if you have time I would mind if you could check that everything still works.

Cargo.toml
```diff
@@ -23,6 +23,14 @@ static_assertions = "1.1"
 version = "0.7.7"
 features = ["critical-section-single-core"]
 
+[dependencies.fdcan]
+# git = "https://github.com/stm32-rs/fdcan.git"
```
kevswims, line 27: Probably don't want to merge with these commented bits in here.

src/lib.rs
```diff
@@ -62,8 +62,8 @@ pub use crate::stm32::interrupt;
 
 pub mod adc;
 pub mod bb;
-// pub mod can;
-pub mod fdcan;
+pub mod can;
+// pub mod fdcan;
```
kevswims, line 66: Remove commented code?

kevswims (changes requested):
I made some changes to the example itself to get this going on my hardware and I was able to get the the two channels on my board to talk to each other. Functionally, I think this is good to go in.

I can put up a follow-on PR that cleans up the example to allow running it either on physical hardware or in loopback mode for testing. I should be able to get that up in a day or two.

PR #58, opened 2023-02-12
Setup defmt and probe run

kevswims:
Setup probe-run configuration for easier running of examples

Update vscode settings for rust analyzer

Add panic example for testing defmt panic logging setup


luctius:
Thank you for this.

I created a separate commit for the fixing of the formatting because I messed there...

Issue #37, opened 2021-11-22
Different clocks for APB1/APB2


burrbull:
Looks like this was copied from stm32g0xx-hal.
https://github.com/stm32-rs/stm32g4xx-hal/blob/d30e860ab5ab6916c3d48ec4c0fe635f2984dd5f/src/rcc/mod.rs#L26-L29

But when G0 has only one APB bus, G4 has 2 which can have different clocks.

PR #59, opened 2023-02-12
Bugfix/apb1 apb2

kevswims:
Fixes https://github.com/stm32-rs/stm32g4xx-hal/issues/37


kevswims:
This should go in after https://github.com/stm32-rs/stm32g4xx-hal/pull/58 which fixes the formatting issues.

PR #68, opened 2023-04-21
Use fugit

usbalbin:
This first started with me trying to implement rtc for this crate by stealing it from stm32g0xx-hal. I noticed there were some differences.

This PR changes the time related types to use fugit to make the crate a bit more consistent with stm32g0xx-hal.

Is this at all something we want?

~(Not at all tested on real hardware yet)~ - Some testing done

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -1,3 +1,6 @@
+/// This code has been taken from the stm32g0xx-hal project and modified slightly to support
+/// STM32G4xx MCUs.
+
 pub use fugit::{
     Duration, ExtU32, HertzU32 as Hertz, HoursDurationU32 as Hour,
     MicrosDurationU32 as MicroSecond, MinutesDurationU32 as Minute, NanosDurationU32 as NanoSecond,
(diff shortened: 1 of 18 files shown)


src/time.rs
```diff
@@ -1,170 +1,114 @@
-use core::ops::{Add, Div};
+pub use fugit::{
```
usbalbin, line 1: The contents of this file is now more or less identical with [the one from stm32g0xx-hal](https://github.com/stm32-rs/stm32g0xx-hal/blob/main/src/time.rs)

src/timer.rs
```diff
@@ -87,14 +87,14 @@ impl CountDownTimer<SYST> {
     }
 }
 
-impl CountDown for CountDownTimer<SYST> {
-    type Time = Hertz;
+impl embedded_hal::timer::CountDown for CountDownTimer<SYST> {
+    type Time = MicroSecond;
```
usbalbin, line 91: I realize Hertz -> MicroSecond is quite a big change, it felt more natural in my mind. However I am fine with reverting that...

usbalbin:
Any thoughts on this? Is this something that is wanted or should I close this PR?

Also please let me know if you want me to change anything :)

no111u3 (approved):

usbalbin:
Thanks a lot :) @no111u3

PR #71, opened 2023-05-20
Rtc

usbalbin:
This depends on #68 (and thus includes its commits for now)

src/rtc.rs is more or less taken from [stm32g0xx-hal](https://github.com/stm32-rs/stm32g0xx-hal/blob/main/src/rtc.rs) and modified slightly to compile for this crate.

Not yet tested on real hardware

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ paste = "1.0"
 bitflags = "1.2"
 vcell = "0.1"
 static_assertions = "1.1"
+fugit = "0.3.5"
 
 [dependencies.cortex-m]
 version = "0.7.7"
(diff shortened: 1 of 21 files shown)


no111u3:
@usbalbin this PR looks outdated and there are more non-necessary stuff.

usbalbin:
> @usbalbin this PR looks outdated and there are more non-necessary stuff.

I am planning on getting back to this at some point early next year. Will remove all unrelated commits then.

jbtheou:
FYI tested on HW with a small change, and works properly for my usecase (VBAT and a supercap)
https://github.com/impulse-internal/stm32g4xx-hal/commit/e99341d30c131abb9af2db117ee68bb81dbcde50

PR #73, opened 2023-05-24
Fix defmt examples

usbalbin:
Resolves #72 for `--features log-rtt,defmt`

Not sure how to do about rtt with non-defmt...

I have tested button, blinky, hello and panic with `--features log-rtt,defmt` and that seems to work. However I am not sure how to test log-itm etc.

Issue #72, opened 2023-05-24
`.defmt` section found, but no version symbol


usbalbin:
I get this when trying to run the examples

```
$ cargo run --example blinky --features stm32g474 -- --chip STM32G474RBTx
Compiling stm32g4xx-hal v0.0.1
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
     Running `probe-run --connect-under-reset target/thumbv7em-none-eabihf/debug/examples/blinky --chip STM32G474RBTx`
(HOST) INFO  flashing program (9 pages / 9.00 KiB)
(HOST) INFO  success!
Error: `.defmt` section found, but no version symbol - check your linker configuration
```

Adding the line `use defmt_rtt as _;` in the blinky example makes the error go away, however there is no sign of the logging output. There is no difference when replacing `info!()` with `error!()`.

---

Adding `use defmt_rtt as _;` to the example/hello (which uses `cortex_m_semihosting::hprintln` to print) gives 

```
Compiling stm32g4xx-hal v0.0.1 (stm32g4xx-hal)
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `probe-run --connect-under-reset target/thumbv7em-none-eabihf/debug/examples/hello --chip STM32G474RBTx`
(HOST) INFO  flashing program (13 pages / 13.00 KiB)
(HOST) INFO  success!
────────────────────────────────────────────────────────────────────────────────
────────────────────────────────────────────────────────────────────────────────
stack backtrace:
   0: __c_m_sh_syscall
(HOST) WARN  call stack was corrupted; unwinding could not be completed
(HOST) ERROR error occurred during backtrace creation: debug information for address 0x80017d4 is missing. Likely fixes:
        1. compile the Rust code with `debug = 1` or higher. This is configured in the `profile.{release,bench}` sections of Cargo.toml (`profile.{dev,test}` default to `debug = 2`)
        2. use a recent version of the `cortex-m` crates (e.g. cortex-m 0.6.3 or newer). Check versions in Cargo.lock
        3. if linking to C code, compile the C code with the `-g` flag

Caused by:
    Do not have unwind info for the given address.
               the backtrace may be incomplete.
(HOST) INFO  device halted without error
```

---

To verify, I have tested https://github.com/knurling-rs/app-template which works as expected.

Am I doing something wrong or is this a bug?

stm32g4xx-hal version: 0.0.1
[...]

PR #75, opened 2023-06-01
Opamp pga-mode

usbalbin:
Blocked on #72

This adds support for using op amp in programmable gain mode and also optionally routing its output direclty to an AD channe(without going to to an external pin). This enables use of opamp with minimal usage of external pins. For example only using 1 pin for non-inverting input:

```rust
let (opamp1, opamp2, opamp3, opamp4, ..) = dp.OPAMP.split(&mut rcc);

// Configure op with pa7 as non-inverting input and set gain to x4
let mut opamp2 = opamp2.pga(
    pa7,
    PgaModeInternal::gain(NonInvertingGain::Gain4),
    //Some(gpioa.pa6),
    Option::<PA6<Analog>>::None, // Do not route output to any external pin, use internal AD instead
);

let mut delay = cp.SYST.delay(&rcc.clocks);
let mut adc = dp
    .ADC2
    .claim(ClockSource::SystemClock, &rcc, &mut delay, true);

// Here we can sample the output of opamp2 as if it was a regular AD pin
let sample = adc.convert(&mut opamp2, stm32g4xx_hal::adc::config::SampleTime::Cycles_640_5);
let millivolts = adc.sample_to_millivolts(sample);
info!("opamp2 thus 4x pa7: {}mV", millivolts);
```

Tested on nucleo-g474re whith stm32g474re(with the commits from #73 for logging to work)

### DIFF
--- a/examples/opamp.rs
+++ b/examples/opamp.rs
@@ -1,0 +1,87 @@
+//! Integrated opamps.
+
+#![no_std]
+#![no_main]
+
+use stm32g4xx_hal::adc::AdcClaim;
+use stm32g4xx_hal::adc::ClockSource;
+use stm32g4xx_hal::gpio::gpioa::*;
+use stm32g4xx_hal::gpio::gpiob::*;
+use stm32g4xx_hal::gpio::Analog;
+use stm32g4xx_hal::opamp::opamp1::IntoPga as _;
+use stm32g4xx_hal::opamp::opamp2::IntoPga as _;
+use stm32g4xx_hal::opamp::NonInvertingGain;
+use stm32g4xx_hal::opamp::PgaModeInternal;
+use stm32g4xx_hal::prelude::*;
+
+use utils::logger::info;
+
+#[macro_use]
+mod utils;
+
+#[cortex_m_rt::entry]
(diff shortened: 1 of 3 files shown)


examples/opamp.rs
```diff
@@ -0,0 +1,93 @@
+//! Integrated opamps.
```
usbalbin, line 1: Example based on the one in the doc comment in src/opamp.rs

src/opamp.rs
```diff
@@ -35,22 +35,75 @@
...
+//!     let (_opamp2, _pa7, _none) = opamp2.disable();
+//!
 //!     loop {}
 //! }
 //! ```
 
+// TODO: Add support for locking using the `LOCK` bit in `OPAMPx_CSR`
+// TODO: Add support for calibration
+// TODO: The output can not be a Option<PIN> if we want to handle "route to pin vs adc"
+//       in a compile time way. See OPAINTOEN in OPAMPx_CSR
```
usbalbin, line 57: Any thoughts on this?

no111u3 (approved):

PR #77, opened 2023-06-08
Hacky g491 and g4a1 support

usbalbin:
Tried to fix or mostly just disable the parts of the code that prevents this crate from compiling with `stm32g491` or `stm32g4a1`


usbalbin:
Depending on when there is a new release of stm32g4, I think #102 actually gives us proper support for G91 and G4A1

PR #80, opened 2023-09-07
Fix incorrect mapping of DMA channel to DMAMUX channel

usbalbin:
When trying to use the DMA2 peripheral on an STM32G474 I noticed that my buffer did not seem to get written to by the DMA. After some investigation I think I found the issue. From what I can this dma driver does not use the correct mapping from DMA channel to DMAMUX channel for DMA2.

According to RM0440 under "DMAMUX mapping":

For category 3 and category 4 devices:
* DMAMUX channels 0 to 7 are connected to DMA1 channels 1 to 8
* DMAMUX channels 8 to 15 are connected to DMA2 channels 1 to 8

For category 2 devices:
* DMAMUX channels 0 to 5 are connected to DMA1 channels 1 to 6
* DMAMUX channels 6 to 11 are connected to DMA2 channels 1 to 6

---

The fix in this PR is quite an ugly hack but it does try to keep the API as unchanged as possible. It only fixes the problem on cat 3 and 4 devices for now(all but G431 and G441).

I get the feeling that the driver as written today works on the (wrong) assumption that every DMA has its own DMAMUX, hence the `fn mux_ptr()` in `Instance`. However when implementing the `Instance` for DMA1 and DMA2 they both use the same one and only DMAMUX.

Therefore i suppose it would make sense to completely remove the `fn mux_ptr()` from the trait and find some better way than to use my `const IS_DMA1: bool`.

I am open to suggestions :)


usbalbin:
@no111u3 would you mind having a look at this? :)

no111u3 (approved):

usbalbin:
Thank you! :)

PR #83, opened 2023-09-14
Changes to Flash.rs

amcelroy:
- FlashWriter for certain chips requires the PAGE_SIZE_KB to be passed in. Some chips are single bank with 4096kB page size and some chips are dual bank with 2048kB page size. It seems this should be an option easily configured by the user.
- Can unlock the Options register (at your own risk)
- Changed Flash write to write 2 32-bit words as per the manual.
- Can write arbitrary array lengths to flash. If data is < 8 bytes (2 32-bit words), the data is padded with 0xFF.
- page_erase was fixed. Writing to the pnb() register was wiping the per() bit. Changed register access from write -> modify
- Verify now checks the 2 32-bit words.


src/flash.rs
```diff
@@ -224,22 +255,47 @@ impl<'a, const SECTOR_SZ_KB: u32> FlashWriter<'a, SECTOR_SZ_KB> {
         // Unlock Flash
         self.unlock()?;
 
-        for idx in (0..data.len()).step_by(2) {
+        for idx in (0..data.len()).step_by(8) {
```
systec-ms, line 258: IMHO a comment like the following would be nice:
```suggestion
        // According to RM0440 Rev 7, "It is only possible to program double word (2 x 32-bit data)"
        for idx in (0..data.len()).step_by(8) {
```

systec-ms (commented):
I have tested the implementation with my `stm32g491re` and it works fine.
The only issue I had is attributable to a problem in the PAC.

The `stm32g491re` is a category 4 device with 512 Kbytes flash and a page size of 2048.
Consequently, 256 pages must addressable but according to the svd-file/`PAC` the page number (`pnb`)  bits of the `FLASH_CR` is 7 bit width.
The 7 bit width is correct for category 3 and category 2 devices, but not for category 4 devices where 8 bits would be proper (see RM0440 Rev 7 P. 180 (Embedded Flash memory (FLASH) for category 4 devices) `FLASH_CR` register).

This shouldn't be an issue of your implementation, and should be resolved when this is fixed in [svd2rust](https://github.com/stm32-rs/stm32-rs/).

src/flash.rs
```diff
@@ -224,22 +255,47 @@ impl<'a, const SECTOR_SZ_KB: u32> FlashWriter<'a, SECTOR_SZ_KB> {
...
             self.valid_address(offset + idx as u32)?;
+            // Check the ending address to make sure there is no overflow
+            self.valid_address(offset + 8 + idx as u32)?;
 
-            let write_address = (FLASH_START + offset + idx as u32) as *mut u16;
+            let write_address1 = (FLASH_START + offset + idx as u32) as *mut u32;
+            let write_address2 = (FLASH_START + offset + 4 + idx as u32) as *mut u32;
+
+            let mut word1: u32 = 0;
+            let mut word2: u32 = 0;
```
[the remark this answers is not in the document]
amcelroy: No objections

src/flash.rs
amcelroy: No objections

amcelroy (commented):
Thank you for the review!

systec-ms:
Thanks, for the quick response ;).

Additionally, I have two notes from my colleague:

> For me this is a bit to much intelligence/"magic" within the driver. In my opinion we should check whether the user passes data to us whose length is a multiple of 8 and return an error, if this is not the case.
> Background:
> Imagine, a user doesn't know, that we only can write multiples of 8. This user passes blocks of 4 bytes (this is valid for a lot of MCUs). The first write succeeds, because you added the missing 4 bytes. The next write will fail, because writing 0xFF is internally another state than a deleted flash block and the MCU's flash routine will return an error, if the user tries to write his next block of 4 bytes.
> We should inform the user that he does not fulfill the boundary conditions as soon as he writes the first time (via an error). This is because the error in the second write is de facto only a subsequent error.

My colleague is also not a huge fan of the possible endless `while` loops (he would like a timeout) but this isn't an issue of this specific pull request and also part of all STM32-HALs as far as I can tell.

src/flash.rs
```diff
@@ -224,22 +255,48 @@ impl<'a, const SECTOR_SZ_KB: u32> FlashWriter<'a, SECTOR_SZ_KB> {
...
+
+            // Check if there is enough data to make 2 words, if there isn't, pad the data with 0xFF
+            if idx + 8 > data.len() {
+                let mut tmp_buffer = [255 as u8; 8];
+                for jdx in idx..data.len() {
+                    tmp_buffer[jdx] = data[idx + jdx];
+                }
+                let tmp_dword = u64::from_le_bytes(tmp_buffer);
+                word1 = tmp_dword as u32;
+                word2 = (tmp_dword >> 32) as u32;
```
systec-ms, line 279: A note from my colleague:

> For me this is a bit to much intelligence/"magic" within the driver. In my opinion we should check whether the user passes data to us whose length is a multiple of 8 and return an error, if this is not the case.
> Background:
> Imagine, a user doesn't know, that we only can write multiples of 8. This user passes blocks of 4 bytes (this is valid for a lot of MCUs). The first write succeeds, because you added the missing 4 bytes. The next write will fail, because writing 0xFF is internally another state than a deleted flash block and the MCU's flash routine will return an error, if the user tries to write his next block of 4 bytes.
> We should inform the user that he does not fulfill the boundary conditions as soon as he writes the first time (via an error). This is because the error in the second write is de facto only a subsequent error.

amcelroy: Yes, I see y'alls perspective. On a previous MCU I worked with, as long as the alignment was fine, a flash bits could go from 1 -> 0, so padding with 0xFF didn't cause an issue and made writing data easier since the user didn't need to worry about data length, only alignment of the start address and that the flash block was already empty. 

Let me write some test cases to cover this and see what happens. If there are any issues with this test I will add a check to ensure incoming arrays are divisible by 8 and return an error (or add a new one) that makes this condition clear. 

Thank you for the feedback.

amcelroy: I did a bit more testing and recognize your colleagues complaint. However, I still would like writing data to be easy in most cases. A hopeful solution is adding a flag to `FlashWriter.write` called `force_data_padding`, along with function comments, that makes it clear that data padding is going on.

examples/flash_with_rtic.rs
```diff
@@ -0,0 +1,220 @@
...
+        const FLASH_SPACING: u32 = 16; // Separate flash writes by 16 bytes
+
+        flash_writer.erase(0x1FC00, 128).unwrap(); // Erase entire page
+
+        for i in 0..6 {
+            match i {
+                0 => {
+                    // This test should fail, as the data needs to be divisible by 8 and force padding is false
+                    let result = flash_writer.write(0x1FC00 + i * FLASH_SPACING, &one_byte, false);
+                    assert!(result.is_err());
```
no111u3, line 99: Might be add logs to more verbose example

examples/flash_with_rtic.rs
```diff
@@ -0,0 +1,220 @@
...
+        let mut flash_writer = flash.writer::<2048>(FlashSize::Sz256K);
+        const FLASH_SPACING: u32 = 16; // Separate flash writes by 16 bytes
+
+        flash_writer.erase(0x1FC00, 128).unwrap(); // Erase entire page
+
+        for i in 0..6 {
+            match i {
+                0 => {
+                    // This test should fail, as the data needs to be divisible by 8 and force padding is false
+                    let result = flash_writer.write(0x1FC00 + i * FLASH_SPACING, &one_byte, false);
```
no111u3, line 98: Use constant instead magic numbers

src/flash.rs
```diff
@@ -217,29 +253,60 @@ impl<'a, const SECTOR_SZ_KB: u32> FlashWriter<'a, SECTOR_SZ_KB> {
...
-            let write_address = (FLASH_START + offset + idx as u32) as *mut u16;
+            let write_address1 = (FLASH_START + offset + idx as u32) as *mut u32;
+            let write_address2 = (FLASH_START + offset + 4 + idx as u32) as *mut u32;
+
+            let word1: u32;
+            let word2: u32;
+
+            // Check if there is enough data to make 2 words, if there isn't, pad the data with 0xFF
+            if idx + 8 > data.len() {
+                let mut tmp_buffer = [255 as u8; 8];
```
no111u3, line 283: should be `255u8`?

no111u3: also, why this buffer length use?

amcelroy: That buffer is used to pad non-divisible-by-8 data if the user wants an "easy" way to write data to flash. The STM32G4 requires 8 bytes to be written at a time. The user can use the `force_data_padding` flag so that data that isn't divisible by 8 is buffered into `tmp_buffer` and appropriately padded on the fly. 

I can change it to 255u8 if that is preferred.

no111u3: Great, also don't forget remove TODO for flash register structure.

amcelroy: Ok, I think it is complete, thank you for reviewing!

no111u3 (approved):

PR #84, opened 2023-10-03
Added the possibility to configure PWM as open drain output

NandoBongers:


src/pwm.rs
```diff
@@ -561,12 +561,26 @@ pins! {
         CH1(ComplementaryDisabled): [
             PA2<Alternate<AF9>>,
             PB14<Alternate<AF1>>,
-            PF9<Alternate<AF3>>
+            PF9<Alternate<AF3>>,
+
+            #[cfg(feature = "pwm-open-drain")]
+            PA2<AlternateOD<AF9>>,
```
usbalbin, line 567: What about changing the pin! macro so that we only need to specify pin and alternate function here? The macro would then generate the things both for `Alternate` and `AlternateOD`. This would avoid some repetition, not sure if it is worth the extra complexity though. Just a though...

So instead of having to specify

```rust
PA2<Alternate<AF9>>,
PA2<AlternateOD<AF9>>,
```
, something like this would suffice
```rust
 PA2: AF9,
```

NandoBongers: I think that's a good suggestion. I will pick this up after the Christmas holidays. Thank you.

NandoBongers: I changed the pin macro as you suggested.

usbalbin: Great! :)

src/pwm.rs
```diff
@@ -431,282 +433,308 @@ pins_tuples! {
...
+        #[cfg(feature = "pwm-open-drain")]
+        impl FaultPins<$TIMX> for $PINBRK<AlternateOD<$ALTBRK>> {
+            const INPUT: BreakInput = BreakInput::BreakIn;
+        }
+    };
+
+    (BRK2: #[ $( $pmeta:meta )* ] $PINBRK2:ident, $ALTBRK2:ty, $TIMX:ty) => {
+        $( #[ $pmeta ] )*
+        impl FaultPins<$TIMX> for $PINBRK2<Alternate<$ALTBRK2>> {
+            const INPUT: BreakInput = BreakInput::BreakIn;
```
usbalbin, line 475:
```suggestion
            const INPUT: BreakInput = BreakInput::BreakIn2;
```

usbalbin:
What about going as far as [this](https://github.com/NandoBongers/stm32g4xx-hal/pull/1)

PR #86, opened 2023-10-27
Fix unused borrow warning in gpio `set_speed`

liamkinne:


no111u3:
Resolved by better way

PR #92, opened 2023-11-29
CI

usbalbin:
* Add g491 and 4a1 to ci, check examples, check with different feature sets

* Remove unused feature embedded-can-03

* Remove feature unstable-defmt from ci

* Add clippy to ci

* Fix warning


no111u3 (approved):

usbalbin:
I *might* have resolved those remaining warnings somewhere in my [messy branch](https://github.com/stm32-rs/stm32g4xx-hal/compare/main...usbalbin:stm32g4xx-hal:hrtim2) with too many changes. Will try to take a look later..

usbalbin:
Btw, https://github.com/DoumanAsh/cortex-m-log seems to be not so active. Not sure how many users use that vs rtt or semihosting

usbalbin:
As I see it, we have atleast two options:
1. Change utils::logger::init() to return the logger instance (if or `()` where not applicable) and add $logger as the first argument to *all* println impl's to make the syntax identical
2. Not use cortex-m-log in our examples

@no111u3 do you have any suggestions? :)

usbalbin:
> As I see it, we have atleast two options:
> 
> 1. Change utils::logger::init() to return the logger instance (if or `()` where not applicable) and add $logger as the first argument to _all_ println impl's to make the syntax identical
> 2. Remove `cortex-m-log` as an option for our examples
> 
> @no111u3 do you have any suggestions? :)

It was possible to work around the issue by making println use log::info which does not need a logger instance (assuming logger::init has been called.

This PR should now be ready for review :)

no111u3 (approved):

PR #96, opened 2023-11-30
WIP: HRTIM

usbalbin:
Work in progress.

**Currently requires https://github.com/stm32-rs/stm32-rs/pull/860** , thus all the CI errors

Thus this(or some future version of this) should probably not be merged until after #102

TODO:
* [ ] Traitify `PreloadSource`, `MasterPreloadSource`
* [ ] Update examples for new traitified changes
    * [x] hrtim/adc-trigger
* [ ] Make sure the timers are started in the examples
* [ ] Go through the examples once more in case of any last minute experiment leftovers


examples/hrtim/eev-comp.rs
```diff
@@ -0,0 +1,143 @@
...
+        .HRTIM_TIMA
+        .pwm_advanced(pin_a, &mut rcc)
+        .prescaler(prescaler)
+        .eev_cfg(EevCfgs::default().eev4(EevCfg::default()))
+        .period(0xFFFF)
+        .finalize(&mut hr_control);
+
+    out1.enable_rst_event(&cr1); // Set low on compare match with cr1
+    out1.enable_rst_event(eev_input4);
+    out1.enable_set_event(&timer); // Set high at new period
```
usbalbin, line 128: @no111u3  sort of related to #98...

I have been experimenting a bit with different ways to specify different types of event sources.

Here an event is specified by passing a reference to the thing to listen to. In this case:
* reset output on `compare register 1` match
* reset output on `external event 4` signal
* set output on timer period event

The nice thing about this is that there is no way to specify the wrong event source by accident. Also there is no need to figure out what enum and variant to to use, just pass in the actual object(or reference to it).

However this requires the event source to exist before the listener which might make things more complicated?

no111u3: @usbalbin thank for this great example. Yesterday I've merged some changes but it will update in some next sessions (for every timers also). So I think about correct and wrong source for event, it might be provided by internal mark field (I'm not sure that rust can do it).

> However this requires the event source to exist before the listener which might make things more complicated?

Following the RM on G4 (and other) family:
>Note: The clock of the slave timer or ADC must be enabled prior to receive events from the master timer, and must not be changed on-the-fly while triggers are received from the master timer.

That means consumer must be created first, I've checked it on my g474 — violation rule makes ADC crazy.

examples/hrtim/adc-trigger.rs
```diff
@@ -0,0 +1,150 @@
...
+        .finalize(&mut hr_control);
+
+    out1.enable_rst_event(&cr1); // Set low on compare match with cr1
+    out2.enable_rst_event(&cr1);
+
+    out1.enable_set_event(&timer); // Set high at new period
+    out2.enable_set_event(&timer);
+
+    out1.enable();
+    out2.enable();
```
usbalbin, line 136: We need to sert cr3 and cr4 since this is what triggers the AD readings

examples/hrtim/adc-trigger.rs
```diff
@@ -0,0 +1,150 @@
...
+        .wait_for_calibration();
+    let mut hr_control = hr_control.constrain();
+    let (timer, (cr1, _cr2, _cr3, _cr4), (mut out1, mut out2)) = dp
+        .HRTIM_TIMA
+        .pwm_advanced((pin_a, pin_b), &mut rcc)
+        .prescaler(prescaler)
+        .period(0xFFFF)
+        // alternated every period with one being
+        // inactive and the other getting to output its wave form
+        // as normal
```
usbalbin, line 126: This comment does not belong here

examples/hrtim/adc-trigger.rs
```diff
@@ -0,0 +1,150 @@
...
+    //        .  50%          .
+    //         ------          ------
+    //out1    |      |        |      |
+    //        |      |        |      |
+    // --------      ----------      --------
+    let (hr_control, ..) = dp
+        .HRTIM_COMMON
+        .hr_control(&mut rcc)
+        .enable_adc_trigger1_source(Adc13Trigger::TimACmp3)
+        .enable_adc_trigger1_source(Adc13Trigger::TimACmp4)
```
usbalbin, line 116: If this enable method was instead defined on `HrPwmControl` then the variables `cr3` and `cr4` could be passed directly to that method without the need for the enum.

examples/hrtim/adc-trigger.rs
```diff
@@ -0,0 +1,150 @@
...
+    let pa0 = gpioa.pa0.into_analog();
+
+    info!("Setup Adc1");
+    let mut adc = dp
+        .ADC1
+        .claim(ClockSource::SystemClock, &rcc, &mut delay, true);
+
+    adc.set_external_trigger((
+        adc::config::TriggerMode::RisingEdge,
+        ExternalTrigger12::Hrtim_adc_trg_1,
```
usbalbin, line 80: Here is yet another enum... What if this argument was a struct that could only be obtained by calling `hr_control.enable_adc_trigger1_source` or something

examples/hrtim/adc-trigger.rs
usbalbin: @no111u3 what do you think about something like this d8d2b92965d822afcdd5c9ac41ea101c28fa41b5 to avoid enums? (Will probably update the examples at some point)

no111u3: Hmm, I think it better to use one work variant instead of any workless.

usbalbin:
I have now gotten rid of a few enums in favor of traits to force the user to pass a reference to the actual type instead of an enum variant. See `examples/adc-trigger.rs` for how that looks like(this is the only example updated so far, it is not yet tested).

For anyone wanting to give it a try, I am using usbalbin/stm32-rs@ab9082b55029f03cedcf54495dd3a9eb83a59634 until the next stm32-rs/stm32-rs release

usbalbin:
@liamkinne or anyone, I am using this or rather [yet another branch](https://github.com/usbalbin/stm32g4xx-hal/tree/hrtim3) at work and I would at some point like to merge this or some future version of it.

Since this is starting to turn into quite big set of changes, do you have any preferences for how it should be done? It is all just one single peripheral device so of course there are interdependencies. So splitting things up into multiple PRs, while possible, might be less fun after a certain point.

usbalbin:
Also I am using some of the not yet released features from the pac so depending a bit on when the next pac will be released I would prefer to at least wait for that.

liamkinne:
I'm not familiar enough with the peripheral to know if it would be easy to break down. Happy to help in reviewing when that time comes.

liamkinne:
@usbalbin was just thinking about this for an upcoming project. Do you think it makes sense do this as a separate `hrtim` crate like `fdcan` so it can be shared between all of the families that have this peripheral?

usbalbin:
Not sure how big the difference is between HRTIM in g4x4 vs f334 vs h7, but sure why not :)

liamkinne:
According to this (section 1.4): https://www.st.com/resource/en/application_note/an4539-hrtim-cookbook-stmicroelectronics.pdf

There are only a few differences.

tbh I would still go ahead an when it's ready get this merged into G4-hal and later look at making it generic.

usbalbin:
> According to this (section 1.4): https://www.st.com/resource/en/application_note/an4539-hrtim-cookbook-stmicroelectronics.pdf
> 
> There are only a few differences.

Thats seems promising :)

> tbh I would still go ahead an when it's ready get this merged into G4-hal and later look at making it generic.

I currently depend on yet to be released pac features. However if that takes too long then perhaps I will try to make it work with the current version. Or ripp out the affected changes in my future PR.

usbalbin:
I will close this in favor of #146

PR #104, opened 2023-12-17
Flash changes

usbalbin:
~~(This PR contains too many unrelated commits, will clean that up)~~ - Done

Noticed some weird things when trying to use the flash on a nucleo-g474. Some of it related to dual bank but not not all of it...

At some point i got the feeling that reading during the verify-thing in `page_erase` and `write` gave weird values back as if there were some race condition or maybe wrong data width or something... Will try to find the problem.

### DIFF
--- a/examples/flash_with_rtic.rs
+++ b/examples/flash_with_rtic.rs
@@ -7,7 +7,7 @@
 
 mod utils;
 
-#[rtic::app(device = stm32g4xx_hal::stm32g4::stm32g474, peripherals = true)]
+#[rtic::app(device = stm32g4xx_hal::stm32, peripherals = true)]
 mod app {
     use crate::utils::logger;
     use stm32g4xx_hal::flash::{FlashExt, FlashSize, FlashWriter, Parts};
@@ -25,18 +25,6 @@ mod app {
     #[local]
     struct Local {}
 
-    fn compare_arrays(a: &[u8], b: &[u8]) -> bool {
-        if a.len() != b.len() {
-            return false;
-        }
-        for i in 0..a.len() {
-            if a[i] != b[i] {
-                return false;
-            }
-        }
-        true
-    }
-
     #[init]
     fn init(cx: init::Context) -> (Shared, Local, init::Monotonics) {
         crate::utils::logger::init();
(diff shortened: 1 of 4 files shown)


examples/flash_with_rtic.rs
```diff
@@ -45,258 +33,74 @@ mod app {
         let cp = cx.core;
 
         let rcc = dp.RCC.constrain();
-        let mut pll_config = stm32g4xx_hal::rcc::PllConfig::default();
-
-        // Sysclock is based on PLL_R
-        pll_config.mux = stm32g4xx_hal::rcc::PLLSrc::HSI; // 16MHz
-        pll_config.n = stm32g4xx_hal::rcc::PllNMul::MUL_32;
-        pll_config.m = stm32g4xx_hal::rcc::PllMDiv::DIV_2; // f(vco) = 16MHz*32/2 = 256MHz
-        pll_config.r = Some(stm32g4xx_hal::rcc::PllRDiv::DIV_2); // f(sysclock) = 256MHz/2 = 128MHz
```
usbalbin, line 54: Correct me if I am wrong but the clock settup is not really needed for this example, right? Thought I'd remove all but the essentials to give more focus on the flash related things

src/flash.rs
```diff
@@ -48,21 +47,22 @@ pub enum FlashSize {
     Sz1M = 1024,
 }
 impl FlashSize {
-    const fn kbytes(self) -> u32 {
+    const fn bytes(self) -> u32 {
```
usbalbin, line 50: Changed tthis to bytes since, from what I can tell really returns a value in units of bytes

src/flash.rs
```diff
@@ -48,21 +47,22 @@ pub enum FlashSize {
...
         SZ_1K * self as u32
     }
 }
 
-pub struct FlashWriter<'a, const SECTOR_SZ_KB: u32> {
+pub struct FlashWriter<'a, const SECTOR_SIZE: u32> {
```
usbalbin, line 55: Same here

usbalbin:
I just had a sneak peak at [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/flash/operations.rs) which puts memory fences in the `program` function to prevent the compiler from doing bad things. That is likely the way to actually solve the problem.

usbalbin:
I will try to re-visit this at a later time

PR #108, opened 2024-01-01
Allow log imports to go unused for examples

liamkinne:
Currently building blinky on main will fail to build because of `#![deny(warnings)]` causing unused imports to error.

This PR fixes that by allowing the log imports to go unused. Works for both `defmt` and `log`.


liamkinne:
It appears there are a lot of errors being thrown by warnings... should I go hunt all of these down in one PR or take them on piecemeal?

no111u3 (approved):

PR #113, opened 2024-01-04
More ergonomic opamp API, add lock and pga functionality

stabler:
Here's an attempt to improve the ergonomics of the opamp API. This PR also adds the remaining PGA modes.

Key changes:
* Moved all the opamp traits up a level, so it's the same trait for all opamps (rather than a separate trait definition for each channel).
* Removed the `Option<>` from the output interface, and instead implemented traits on the output pin directly, or a new type `InternalOutput` that represents the case where the opamp is directly connected to the ADC.
* Implemented a `Locked<_>` struct. Opamps can be moved from any of the implemented modes (`Pga`, `OpenLoop`, `Follower`) into the `Locked` state, at which point the lock bit is set and you cannot modify or disable the opamp (but you can pass it into the ADC, if appropriately configured)
* Added remaining Pga functionality.
* Consolidated the ADC implementations. As far as I can tell, the macros `adc_op_pga` and `adc_op_follower` were intended to work around limitations of the opamp (certain opamps can only operate in certain modes), but I can't see anything about that in the datasheet or reference manual. Maybe this is an artifact from copy-paste from another chip?

Things I think are good:
* I think this makes the interface more ergonomic- particularly when switching between modes (no more option passed back that needs to be unwrapped)
* Easy to prevent the user from passing the opamp into the ADC when it isn't configured to output to the internal channel
* Easy to support the locked state

Things I don't love
* ~There's an unwrap in some `disable()` and all `disable_output()` functions. This is safe because those functions are only implemented on opamps that contain an output pin, but it relies on the module being implemented correctly.~
* I think there's probably a way to reduce the code duplication in the macro and better use traits, but I couldn't find a way to do it that made the compiler happy

Let me know what you think! @usbalbin I know you worked on this code recently, so you may have opinions!


stabler:
Turns out I didn't need the unwrap after all (ff07d75)

src/opamp.rs
```diff
@@ -172,161 +238,213 @@ macro_rules! opamps {
...
+                    unsafe fn _reset() {
+                        (*crate::stm32::OPAMP::ptr()).[<$opampreg _csr>].reset()
+                    }
 
-                    #[allow(unused_imports)]
-                    use crate::gpio::gpiob::*;
+                    /// Disable the output (connect the opamp to the internal ADC
+                    /// channel). Used by internal implementations.
+                    unsafe fn _disable_output() {
+                        (*crate::stm32::OPAMP::ptr()).[<$opampreg _csr>].write(|w|
```
usbalbin, line 250: Is this and `_enable_output` supposed to overwrite(not modify) the entire csr register?

stabler: No, good catch, fixed here: 13d4631

I made a small change to consolidate some of the code too. It results in an additional register write when constructing the opamps, but that seems like a negligible penalty.

src/opamp.rs
```diff
@@ -370,46 +488,76 @@ macro_rules! opamps {
...
+                        (*crate::stm32::OPAMP::ptr())
+                            .[<$opampreg _csr>]
+                            .write(|csr_w|
+                                csr_w
+                                    .vp_sel()
+                                    .$input_mask()
+                                    .vm_sel()
+                                    .output()
+                                    .opaintoen()
+                                    .variant(OPAINTOEN_A::OutputPin)
```
usbalbin: Not sure if it would be worth it, but what if we had a trait with an associated const `OPAINTOEN_A` that was implemented for external and InternalOutput types?

usbalbin:
I'm just a user, but from what I have seen so far this looks great. I really like not having 6 different modules with their own types. I also like not having to use the `Option`'s for the output :)

Awesome :)

stabler:
For those interested, here's a branch where I'm working on adding the remaining Pga modes:
https://github.com/stabler/stm32g4xx-hal/commit/861a3e77aae5c4d2330a31c1ed80200d956e42cd

It's a bit heavy handed right now (separate structs for each Pga variant), so I'm going to see if I can consolidate a bit (and test it!) before I put a PR up

src/opamp.rs
stabler: Take a look at the other change I made on 13d4631 and let me know what you think. I'll play with the trait idea.

usbalbin: Taking that one step further, what if all follower, pga, etc defaulted to internal only? Then the user could enable the output builder style. That would probably compose well with optional inverting inputs etc.

stabler: I thought about that but I think complexity will increase a lot (e.g. we would need to store various configurations of outputs attached to the struct). I don't think that makes sense given the limited surface area of the API. I think this is probably the 'correct' way to do it though.

stabler:
Ok I cleaned up the Pga implementation (a lot!), so I pulled it into this PR. I haven't tested all of this code on hardware yet (hoping to do that this weekend), but I'm feeling pretty good about the Pga implementation.

stabler:
Nice compile-time error if you use the wrong input pin in the more complex Pga modes:
image

src/opamp.rs
```diff
@@ -28,81 +30,229 @@
...
+/// Internal trait implementing the low level register write used to
+/// configure the PGA.
+trait ConfigurePgaReg<Opamp, NonInverting>
+where
+    Opamp: LookupPgaGain,
+{
+    /// Type of the associated vinm0 input.
+    type vinm0;
+    /// Type of the associated vinm1 input.
+    type vinm1;
```
usbalbin, line 202: CI seems unhappy with these names.

What about something like this? :)

```suggestion
    /// Type of the associated vinm0 input.
    type Vinm0;
    /// Type of the associated vinm1 input.
    type Vinm1;
```

stabler: Thanks @usbalbin  - 489c268

I don't have the hardware set up to test this any more, and likely won't be able to get back to it in the near future, but it was working well when I put it down.

usbalbin (approved):
Thanks a lot :)

PR #116, opened 2024-01-05
Add dma fn read_available and uart-dma-rx example

usbalbin:

PR #119, opened 2024-01-23
DMA: Add read_available

usbalbin:


src/dma/transfer.rs
```diff
@@ -340,9 +340,8 @@ where
     pub fn read_exact(
         &mut self,
         dat: &mut [<PERIPHERAL as TargetAddress<PeripheralToMemory>>::MemSize],
-    ) -> usize {
+    ) {
```
usbalbin, line 343: Since the returned value is always `data.len()` i see no pint in returning it.

usbalbin:
Seems I somehow forgot about #116

PR #123, opened 2024-02-21
Add internal DAC outputs to available opamp inputs.

davidstalnaker:
This is done by adding internal structs for the DAC outputs and passing these into the constructor functions instead of an input pin, e.g. `opamp1.follower(Dac3Out1, Some(gpioa.pa2.into_analog()));`

I've tested that this works at least with opamps 1, 3, 4, and 5 in follower mode on an STM32G473.  Can't verify everything of course but I don't see anything obvious that would break elsewhere, and it at least builds for other devices.

After writing this I now see related work in https://github.com/stm32-rs/stm32g4xx-hal/pull/113.  If I'm reading right, I think this change is completely orthogonal and should merge with that one pretty cleanly.

### DIFF
--- a/src/opamp.rs
+++ b/src/opamp.rs
@@ -96,6 +96,18 @@ pub enum NonInvertingGain {
     Gain64 = 5,
 }
 
+/// Internal output of DAC 3 channel 1.
+pub struct Dac3Out1;
+
+/// Internal output of DAC 3 channel 2.
+pub struct Dac3Out2;
+
+/// Internal output of DAC 4 channel 1.
+pub struct Dac4Out1;
+
+/// Internal output of DAC 4 channel 2.
+pub struct Dac4Out2;
+
 macro_rules! opamps {
     {
         $(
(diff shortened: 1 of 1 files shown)


src/opamp.rs
```diff
@@ -96,6 +96,18 @@ pub enum NonInvertingGain {
...
+
+/// Internal output of DAC 3 channel 2.
+pub struct Dac3Out2;
+
+/// Internal output of DAC 4 channel 1.
+pub struct Dac4Out1;
+
+/// Internal output of DAC 4 channel 2.
+pub struct Dac4Out2;
+
```
usbalbin, line 110: We have the channel types from the [dac module](https://github.com/davidstalnaker/stm32g4xx-hal/blob/opamp-dac/src/dac.rs#L73) if we want to use those. I believe that would force the user to pass in the DAC they actually initialized, reducing the risk of error.

davidstalnaker: I like the idea but I don't think taking ownership of the DAC channel works in this case, for two reasons:

1. The same DAC output can be used as input for multiple opamps (dac3ch1 can go to opamps 1 and 6 simultaneously, verified that this works).
2. The user wouldn't be able to set the DAC's value (or other configuration) after setting up the opamp.  There's some cases where that's ok (if using DMA) but the easiest setup would become impossible.

usbalbin: Passing reference would probably be enough. As long as the OPAMP does not keep the reference, the dac can still be modified and similar which I believe would solve both 1 and 2?

usbalbin: Something similar is done for the [comparator](https://github.com/davidstalnaker/stm32g4xx-hal/blob/opamp-dac/src/comparator.rs#L289).

PR #128, opened 2024-07-27
Fix outstanding clippy warnings

liamkinne:
- Fix feature names for chip variant `stm32g4A1` -> `stm32g4a1` and some typos of `stm32g49a`.
- Fix comment indentation in comments
- Fix missing type in `mem::transmute` use

PR #129, opened 2024-07-30
Add embedded-hal 1.0 trait support

### LINKED ISSUE #118: Move to embedded-hal 1.0
Now that 1.0 is released, is there a plan already to move the HAL to this version?

techmccat:
Fixes #118 

The new traits are not feature gated, instead embedded_hal 0.2 is now exported as hal_02
Traits from embedded-hal 1.0 have replaced the old ones in the prelude

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,8 @@ cortex-m-log = { version = "0.7", features = ["log-integration"] }
 cfg-if = "0.1.10"
 mpu6050 = "0.1.4"
 bme680 = "0.6.0"
+sh1106 = { git = "https://github.com/techmccat/sh1106.git", branch = "hal-1", version = "0.5.0" }
+embedded-graphics = "0.8.0"
 embedded-sdmmc = "0.3.0"
 
 #TODO: Separate feature sets
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,13 +37,13 @@ default-features = false
 features = ["const-fn"]
 version = "0.2.5"
 
-[dependencies.embedded-hal]
-features = ["unproven"]
-version = "0.2.4"
-
-[dependencies.embedded-hal-one]
-version = "1.0.0"
+[dependencies.embedded-hal-old]
 package = "embedded-hal"
+features = ["unproven"]
+version = "0.2.7"
+
+[dependencies.embedded-hal]
+version = "1.0.0"
 
 [dependencies.embedded-dma]
 version = "0.1.2"
@@ -95,7 +95,7 @@ stm32g4a1 = ["stm32g4/stm32g4a1"]
 log-itm = ["cortex-m-log/itm"]
 log-rtt = []
 log-semihost = ["cortex-m-log/semihosting"]
-defmt-logging = ["defmt", "nb/defmt-0-3", "embedded-hal-one/defmt-03", "embedded-io/defmt-03"]
+defmt-logging = ["defmt", "nb/defmt-0-3", "embedded-hal/defmt-03", "embedded-io/defmt-03"]
 
 [profile.dev]
 codegen-units = 1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,10 @@ version = "0.2.5"
 features = ["unproven"]
 version = "0.2.4"
 
+[dependencies.embedded-hal-one]
+version = "1.0.0-rc.3"
+package = "embedded-hal"
+
 [dependencies.embedded-dma]
 version = "0.1.2"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ features = ["unproven"]
 version = "0.2.4"
 
 [dependencies.embedded-hal-one]
-version = "1.0.0-rc.3"
+version = "1.0.0"
 package = "embedded-hal"
 
 [dependencies.embedded-dma]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,13 +12,14 @@ repository = "https://github.com/stm32-rs/stm32g4xx-hal"
 version = "0.0.2"
 
 [dependencies]
-nb = "0.1.1"
+nb = "1"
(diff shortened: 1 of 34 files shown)


examples/adc-continious-dma.rs
```diff
@@ -47,7 +49,10 @@ fn main() -> ! {
     let pa0 = gpioa.pa0.into_analog();
 
     info!("Setup Adc1");
-    let mut delay = cp.SYST.delay(&rcc.clocks);
+    // let mut delay = cp.SYST.delay(&rcc.clocks);
+    let mut delay = DelayFromCountDownTimer::new(
+        Timer::new(dp.TIM6, &rcc.clocks).start_count_down(100u32.millis()),
+    );
```
usbalbin, line 55: Why is this needed? :)

techmccat: I switched out the SYST delay provider because cortex-m doesn't implement the new delay trait yet, it's probably not necessary though

usbalbin:
There appears a bunch of unrelated changes. Since this PR is quite large, would it be possible to keep this PR to only the e-hal 1.0 migration? :)

Also the CI was has just been fixed in #128, which was just merged. Would it be possible to rebase over the latest changes to get better ci results?

techmccat:
I've tried cherry-picking only the hal-1 related changes, suppose I'll put SPI driver improvements and i2c example in separate PRs once this one goes through.

Cargo.toml
```diff
@@ -70,6 +75,8 @@ cortex-m-log = { version = "0.7", features = ["log-integration"] }
 cfg-if = "0.1.10"
 mpu6050 = "0.1.4"
 bme680 = "0.6.0"
+sh1106 = { git = "https://github.com/techmccat/sh1106.git", branch = "hal-1", version = "0.5.0" }
```
usbalbin, line 78: Is there any reason we are not using the regular 0.5.0 from crates.io?

examples/adc-continious-dma.rs
usbalbin: Is there any way to avoid this since I would imagine this would break a lot of users code?

examples/pwm.rs
```diff
@@ -26,7 +27,7 @@ fn main() -> ! {
 
     let mut pwm = dp.TIM1.pwm(pin, 100.Hz(), &mut rcc);
 
-    pwm.set_duty(pwm.get_max_duty() / 2);
+    let _ = pwm.set_duty_cycle(pwm.max_duty_cycle() / 2);
```
usbalbin, line 30: Would it even make sense to use [set_duty_cycle_percent](https://docs.rs/embedded-hal/latest/embedded_hal/pwm/trait.SetDutyCycle.html#method.set_duty_cycle_percent) to set 50% duty?

src/serial/usart.rs
```diff
@@ -349,6 +366,91 @@ macro_rules! uart_shared {
...
+                let usart = unsafe { &(*$USARTX::ptr()) };
+                let mut count = 0;
+
+                while !self.read_ready()? {
+                    core::hint::spin_loop()
+                }
+                while self.read_ready()? && count < buf.len() {
+                    buf[count] = usart.rdr.read().bits() as u8;
+                    count += 1
+                }
```
usbalbin, line 423: Will this ever yield more than 1 byte(assuming f_sys > baud_rate)?

src/time.rs
```diff
@@ -101,17 +101,17 @@ impl U32Ext for u32 {
     }
 }
 
-pub fn duration(hz: Hertz, cycles: u32) -> MicroSecond {
+pub fn duration<const NOM: u32, const DENOM: u32>(hz: Hertz, cycles: u32) -> Duration<u32, NOM, DENOM> {
     let cycles = cycles as u64;
     let clk = hz.raw() as u64;
-    let us = cycles.saturating_mul(1_000_000_u64) / clk;
-    MicroSecond::from_ticks(us as u32)
+    let us = cycles.saturating_mul(DENOM as u64) / clk / NOM as u64;
```
usbalbin, line 107: `let us`: Is this even microseconds any more?

Cargo.toml
techmccat: I was using my fork to test the i2c implementation for e-hal 1.
I also thought I hadn't included this example in the new set of changes but I must've re-added it by mistake at some point.

examples/adc-continious-dma.rs
techmccat: Using the old e-hal 0.2 trait should still work, but the function names are in conflict with the new ones which are exported in the prelude

src/serial/usart.rs
techmccat: Manual says there's an 8 byte FIFO that we could enable

src/time.rs
techmccat: Guess not, it would be duration ticks

examples/pwm.rs
techmccat: Good call, I had forgotten it was a thing

src/serial/usart.rs
usbalbin: Oh, nice

examples/adc-continious-dma.rs
usbalbin: Might be a good idea to see how/if this is solved in for example [stm32-rs/stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal)

techmccat: DelayNs has been implemented in https://github.com/rust-embedded/cortex-m/commit/ec6b3ba49db60b23885921d238d3a9e86cf3e9e9 but it hasn't been included in a release yet

usbalbin:
@techmccat I have rebased this and made an attempt at fixing https://github.com/stm32-rs/stm32g4xx-hal/pull/129#discussion_r1697442683 in #172 . I will close this in favor of #172 please reopen if you disagree.

Thanks a lot for this!

PR #131, opened 2024-08-03
Fix `PLLSrc` enum not matching capitalisation conventions

liamkinne:
`PLLSrc` -> `PllSrc`

### DIFF
--- a/examples/flash_with_rtic.rs
+++ b/examples/flash_with_rtic.rs
@@ -48,7 +48,7 @@ mod app {
         let mut pll_config = stm32g4xx_hal::rcc::PllConfig::default();
 
         // Sysclock is based on PLL_R
-        pll_config.mux = stm32g4xx_hal::rcc::PLLSrc::HSI; // 16MHz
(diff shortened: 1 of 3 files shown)


liamkinne:
@usbalbin not sure if you have write permissions yet, but here's another PR to look at. Thanks.

usbalbin:
> @usbalbin not sure if you have write permissions yet, but here's another PR to look at. Thanks.

I do not, as far as I know.

LGTM! :)

However I would (unless this does that) prefer to fix the CI errors first

liamkinne:
@usbalbin the CI errors are all to do with typos which I've fixed in #128. Still waiting on getting that merged as well.

usbalbin:
Great!

However there are, as far as I can tell, still [some clippy warnings](https://github.com/stm32-rs/stm32g4xx-hal/actions/runs/10124403903/job/27998709804?pr=128) left for #128.

usbalbin (approved):
Thanks! :)

PR #132, opened 2024-08-24
Add CORDIC Support

AdinAck:
Create a new HAL component for using the CORDIC co-processor.

The interface permits configuration of all permutations of argument/result size and count, function, scale, and precision.

Configurations are represented by type-states and as such are compile-time validated.

# Todo

I haven't added an interface for DMA (yet).

# Example

As far as I can tell, examples are broken, so I only included a minimal example in the module-level docs. Here is that example:

```rust
#[entry]
fn main() -> ! {
    let pwr = dp.PWR.constrain().freeze();
    let mut rcc = dp.RCC.freeze(Config::hsi(), pwr);

    let cordic = dp
        .CORDIC
        .constrain(&mut rcc)
        .freeze::<Q15, Q31, SinCos, P60>(); // 16 bit arguments, 32 bit results, compute sine and cosine, 60 iterations

    cordic.start(I1F15::from_num(-0.25 /* -45 degreees */));

    let (sin, cos) = cordic.result();

    info!("sin: {}, cos: {}", sin.to_num::<f32>(), cos.to_num::<f32>());

    loop {}
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ bitflags = "1.2"
 vcell = "0.1"
 static_assertions = "1.1"
 fugit = "0.3.5"
+fixed = "1.28.0"
 
 [dependencies.cortex-m]
 version = "0.7.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ bitflags = "1.2"
 vcell = "0.1"
 static_assertions = "1.1"
 fugit = "0.3.5"
-fixed = "1.28.0"
+fixed = { version = "1.28.0", optional = true }
 
 [dependencies.cortex-m]
 version = "0.7.7"
@@ -90,6 +90,7 @@ log-itm = ["cortex-m-log/itm"]
 log-rtt = []
 log-semihost = ["cortex-m-log/semihosting"]
 defmt-logging = ["defmt"]
+cordic = ["dep:fixed"]
 
 [profile.dev]
 codegen-units = 1
@@ -109,4 +110,4 @@ required-features = ["stm32g474"]
 
 [[example]]
 name = "cordic"
-required-features = ["defmt"]
+required-features = ["cordic"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,3 +106,7 @@ lto = true
 [[example]]
 name = "flash_with_rtic"
 required-features = ["stm32g474"]
+
+[[example]]
+name = "cordic"
+required-features = ["defmt"]
--- a/src/cordic.rs
+++ b/src/cordic.rs
@@ -64,30 +64,57 @@ pub mod types {
 
     pub use fixed::types::{I1F15, I1F31};
 
-    /// Trait for newtypes to represent CORDIC argument or result data.
-    pub trait DataType: Fixed {
+    /// q1.15 fixed point number.
+    pub struct Q15;
+    /// q1.31 fixed point number.
+    pub struct Q31;
+
+    /// Extension trait for fixed point types.
+    pub trait Ext: Fixed {
+        /// Type-state representing this type.
+        type Repr: DataType<Inner = Self>;
+
         /// Convert to bits of the register width,
         fn to_register(self) -> u32;
         /// Convert from bits of the register width,
         fn from_register(bits: u32) -> Self;
     }
 
-    impl DataType for I1F31 {
-        fn to_register(self) -> u32 {
-            self.to_bits() as u32
-        }
-        fn from_register(bits: u32) -> Self {
-            Self::from_bits(bits as i32)
-        }
-    }
+    impl Ext for I1F15 {
+        type Repr = Q15;
 
-    impl DataType for I1F15 {
         fn to_register(self) -> u32 {
             self.to_bits() as u16 as u32
         }
         fn from_register(bits: u32) -> Self {
             Self::from_bits(bits as u16 as i16)
         }
+    }
+
+    impl Ext for I1F31 {
+        type Repr = Q31;
+
+        fn to_register(self) -> u32 {
+            self.to_bits() as u32
+        }
+
+        fn from_register(bits: u32) -> Self {
+            Self::from_bits(bits as i32)
+        }
+    }
+
+    /// Trait for newtypes to represent CORDIC argument or result data.
+    pub trait DataType {
+        /// Internal fixed point representation.
+        type Inner: Ext<Repr = Self>;
+    }
+
+    impl DataType for Q15 {
+        type Inner = I1F15;
+    }
+
+    impl DataType for Q31 {
+        type Inner = I1F31;
     }
 
     /// Traits and structures related to argument type-states.
@@ -119,8 +146,8 @@ pub mod types {
         }
 
         impls! {
-            (super::I1F31, Bits32),
-            (super::I1F15, Bits16),
+            (super::Q31, Bits32),
+            (super::Q15, Bits16),
         }
     }
 
@@ -153,8 +180,8 @@ pub mod types {
         }
 
         impls! {
-            (super::I1F31, Bits32),
-            (super::I1F15, Bits16),
+            (super::Q31, Bits32),
+            (super::Q15, Bits16),
         }
     }
 }
@@ -182,7 +209,7 @@ pub mod func {
         where
             T: types::DataType,
         {
-            type Signature: super::signature::Property<T>;
+            type Signature: super::signature::Property<T::Inner>;
 
             const COUNT: Count;
         }
@@ -191,7 +218,7 @@ pub mod func {
         where
             T: types::DataType,
         {
-            type Signature = T;
+            type Signature = T::Inner;
 
             const COUNT: Count = Count::One;
         }
@@ -200,7 +227,7 @@ pub mod func {
         where
             T: types::DataType,
         {
-            type Signature = (T, T);
+            type Signature = (T::Inner, T::Inner);
 
             const COUNT: Count = Count::Two;
         }
@@ -334,6 +361,8 @@ pub mod func {
     /// Traits and structures related to the function signature.
     pub mod signature {
         use super::{data_count, reg_count, types};
+        use types::arg::State as _;
+        use types::res::State as _;
 
         type WData = crate::stm32g4::Reg<crate::stm32::cordic::wdata::WDATA_SPEC>;
         type RData = crate::stm32g4::Reg<crate::stm32::cordic::rdata::RDATA_SPEC>;
@@ -341,7 +370,7 @@ pub mod func {
         /// The signature is a property of the function type-state.
         pub trait Property<T>
         where
-            T: types::DataType,
+            T: types::Ext,
         {
             /// Number of register operations required.
             type NReg;
@@ -349,25 +378,25 @@ pub mod func {
             /// Write arguments to the argument register.
             fn write(self, reg: &WData)
             where
-                T: types::arg::State;
+                T::Repr: types::arg::State;
 
             /// Read results from the result register.
             fn read(reg: &RData) -> Self
             where
-                T: types::res::State;
+                T::Repr: types::res::State;
         }
 
         impl<T> Property<T> for T
         where
-            T: types::DataType,
+            T: types::Ext,
         {
-            type NReg = reg_count::NReg<T, data_count::One>;
+            type NReg = reg_count::NReg<T::Repr, data_count::One>;
 
             fn write(self, reg: &WData)
             where
-                T: types::arg::State,
+                T::Repr: types::arg::State,
             {
-                let data = match const { T::RAW } {
+                let data = match const { T::Repr::RAW } {
                     types::arg::Raw::Bits16 => {
                         // $RM0440 17.4.2
                         // since we are only using the lower half of the register,
@@ -384,7 +413,7 @@ pub mod func {
 
             fn read(reg: &RData) -> Self
             where
-                T: types::res::State,
+                T::Repr: types::res::State,
             {
                 T::from_register(reg.read().res().bits())
             }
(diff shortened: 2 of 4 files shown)


usbalbin:
> As far as I can tell, examples are broken [...]

Anyone example in perticular? They goal is that they should work AFAIK...

usbalbin:
BTW, thanks a lot for your contribution! :)

AdinAck:
> > As far as I can tell, examples are broken [...]
> 
> Anyone example in perticular? The goal is that they should all work AFAIK...

It must be something on my end, as CI of course is successfully building the examples... I'm running this command:

```sh
cargo build --examples --features stm32g431,log-rtt,defmt
```

And it's emitting large linker errors, I don't know why!

```
ERROR(cortex-m-rt): The interrupt vectors are missing.
```

> BTW, thanks a lot for your contribution! :)

Of course!

usbalbin:
I think #133 might be the cause of the errors you are seeing

usbalbin:
Does this work?
```bash
cargo build --example blinky --features stm32g431,log-rtt,defmt
```

AdinAck:
Ahah! You're right! Nice, I'll add an example now. Thanks!

AdinAck:
If you merge #133 I'll pull it from upstream and CI should pass :)

usbalbin:
I have never used this peripheral so I do not know what the typical use case looks like.. But say you wanted to change first run one function and then switch to another function.

```rust
// (some sort of pseudo rustish code)

fn do_math_stuff(cordic: &mut Cordic<...>) {
    cordic.do_sin_calc();
    
    cordic.do_sqrt_calc(); // <--- This would require to (atleast temporary) change type which is not allowed
    cordic.change_back_to_sin_mode();
}
```

I assume this would be not so simple to support. Again, not sure how this is typically used and if it at all is something we care about?

AdinAck:
> ...say you wanted to first run one function and then switch to another function.

Since I felt it imperative for the peripheral state to be encoded in the type system, this can be done with the type system:

```rust
type CordicForSin = Cordic<_, _, Sin, _>;
type CordicForSqrt = Cordic<_, _, Sqrt, _>;

fn do_math_stuff(mut cordic: CordicForSin) -> CordicForSin { // cordic must be moved now
    cordic.start(_);
    let _ = cordic.result();

    let mut cordic: CordicForSqrt = cordic.freeze();
    
    cordic.start(_);
    let _ = cordic.result();

    cordic.freeze()
}
```

If something wants to *change the resource* then `&mut` is not enough.

> I assume this would be not so simple to do as is. Again, not sure how this is typically used and if it at all is something we care about?

I did consider two options for enabling highly dynamic usage:
1. With allocation, one could accept an `&mut` to a `Cordic<_, _, dyn func::State, _>`.
2. `Any` type-states could be implemented for each configurable property.

My priority was maximizing safety and performance, but certainly either of these two options can be easily adopted later.

AdinAck:
I don't know what `itm` or `semihost` are but it seems they change the `println!` macro, I'll just add a cfg...

Cargo.toml
```diff
@@ -105,3 +106,7 @@ lto = true
 [[example]]
 name = "flash_with_rtic"
 required-features = ["stm32g474"]
+
+[[example]]
+name = "cordic"
+required-features = ["defmt"]
```
usbalbin, line 112: should not be needed after #134. 

```suggestion
```

(Apparently the macro was only used and tested without formatting args.)

src/cordic.rs
```diff
@@ -0,0 +1,845 @@
...
+/// Cordic co-processor interface.
+#[allow(private_bounds)]
+pub struct Cordic<Arg, Res, Func, Prec>
+where
+    Arg: data_type::arg::State,
+    Res: data_type::res::State,
+    Func: func::State<Arg, Res>,
+    Prec: prec::State,
+{
+    rb: CORDIC,
```
usbalbin, line 701: Just a thought, what if `rb` was `impl BorrowMut`? Then the user could

```rust
type CordicForSin = Cordic<_, _, Sin, _>;
type CordicForSqrt = Cordic<_, _, Sqrt, _>;

fn do_math_stuff(rb: &mut CORDIC) {
    {
        let mut cordic = rb.freeze_not_consuming();
        cordic.start(_);
        let _ = cordic.result();
    } // <--- rb no longer borrowed by cordic, it is free to be reused

    let mut cordic: CordicForSqrt = rb.freeze();
    
    cordic.start(_);
    let _ = cordic.result();
}
```

again, just a thought. Depending how the cordic is typically used, maybe not worth the extra generic arg.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ bitflags = "1.2"
 vcell = "0.1"
 static_assertions = "1.1"
 fugit = "0.3.5"
+fixed = "1.28.0"
```
usbalbin, line 22: `fixed` seems to require rustc 1.79 or newer. What do you think about putting cordic and `fixed` behind a feature?

Cargo.toml
AdinAck: This command fails now:

```sh
cargo build --examples --features stm32g431,log-rtt
```

Looks like the `log-rtt` feature no longer enables the `defmt` feature?

Cargo.toml
AdinAck: Ah yep, I was thinking about that. Sounds good!

src/cordic.rs
AdinAck: So, you nerdsniped me and I've been mulling this over since yesterday.

I just refactored the whole thing to be much stronger in terms of type relationships and enforcements.

As a result, now, adding this dynamic behavior is easy!

```rust
/// Traits and structures for dynamic function operation.
pub mod dynamic {
    use super::{prec, signature, types, Cordic, State};

    /// Any function can be invoked with this type-state.
    pub struct Any;

    /// A Cordic in dynamic mode.
    pub trait Mode<Arg, Res>
    where
        Arg: types::arg::State,
        Res: types::res::State,
    {
        /// Run a function with provided arguments and get the result.
        ///
        /// *Note: This employs the polling strategy.
        /// For less overhead, use static operations.*
        fn run<Func>(&mut self, args: Func::Arguments) -> Func::Results
        where
            Func: State<Arg, Res>;
    }

    impl<Arg, Res, Prec> Mode<Arg, Res> for Cordic<Arg, Res, Any, Prec>
    where
        Arg: types::arg::State,
        Res: types::res::State,
        Prec: prec::State,
    {
        fn run<Func>(&mut self, args: Func::Arguments) -> Func::Results
        where
            Func: State<Arg, Res>,
        {
            use signature::Property as _;

            self.apply_config::<Arg, Res, Func, Prec>();

            args.write(&self.rb.wdata);
            self.when_ready(|cordic| Func::Results::read(&cordic.rb.rdata))
        }
    }
}
```

So your example would look like this:

```rust
fn do_math_stuff(cordic: &mut Cordic<_, _, Any, _>) {
    let sin = cordic.run::<Sin>(_);
    let sqrt = cordic.run::<Sqrt<N0>>(_);
}
```

src/cordic.rs
```diff
@@ -0,0 +1,915 @@
...
+        /// Trait for argument type-states.
+        pub(crate) trait State: super::DataType {
+            /// Raw representation of the configuration
+            /// in the form of a bitfield variant.
+            const RAW: Raw;
+
+            /// Configure the resource to be represented
+            /// by this type-state.
+            #[inline]
+            fn set<const O: u8>(w: crate::stm32::cordic::csr::ARGSIZE_W<O>) {
```
usbalbin, line 106: (Feel free to ignore)

Maybe it's just me but, `O` looks a hole lot like a zero `0` to me

AdinAck: Yeah I just copied the PAC, I think it stands for `Offset`, I'll change it to that.

src/cordic.rs
```diff
@@ -0,0 +1,915 @@
...
+    /// Arctangent of x (primary) and y (secondary).
+    pub struct ATan2;
+    /// Magnitude of x (primary) and y (secondary).
+    pub struct Magnitude;
+    /// Arctangent (primary) and magnitude (secondary) of x (primary) and y (secondary).
+    pub struct ATan2Magnitude;
+    /// Arctangent of x.
+    ///
+    /// This function can be scaled by 0-7.
+
```
usbalbin, line 513: Is there supposed to be a newline here?
```suggestion
```

AdinAck: Nope!

usbalbin (approved):
Thank you very much! This is awesome!

src/cordic.rs
usbalbin: Wow :)

usbalbin:
Is it ok if I merge now, or is there anything else you want to add?

AdinAck:
Thanks, I'm *pretty* sure we're good 😅. I did start thinking about some minor organizational changes but they should be inconsequential.

Maybe let me sit on it for a day, if that's ok with you.

AdinAck:
I decluttered the type signature from the `fixed` madness to just `Q15` or `Q31`.

Something I realized that is really annoying me right now is that an enforcement strategy I came up with to further strengthen the system invariances seems to be impossible to implement!

Currently one could comment out a line of `apply_config()` and the program would still compile:

```rust
fn apply_config<NewArg, NewRes, NewFunc, NewPrec>(&mut self)
    where
        NewArg: types::arg::State,
        NewRes: types::res::State,
        NewFunc: func::Feature<NewArg, NewRes>,
        NewPrec: prec::State,
    {
        self.rb.csr.write(|w| {
            use func::reg_count::arg::State as _;
            use func::reg_count::res::State as _;
            use func::scale::State as _;

            NewArg::set(w.argsize());
            NewRes::set(w.ressize());
            NewPrec::set(w.precision());
            NewFunc::set(w.func());
            // NewFunc::NArgs::set(w.nargs()); whoopsies!
            NewFunc::NRes::set(w.nres());
            NewFunc::Scale::set(w.scale());

            w
        });
    }
```

Nothing about the signature *requires* all states to be set.

So I decided to make `State::set()` return the corresponding type-state, and `apply_config()` returns all type states, which could then be stored in the `Cordic` type instead of `PhantomData` (they're all ZSTs anyway so it would be the same).

```rust
fn apply_config<NewArg, NewRes, NewFunc, NewPrec>(&mut self) -> (NewArf, NewRes, ..)
    where
        NewArg: types::arg::State,
        NewRes: types::res::State,
        NewFunc: func::Feature<NewArg, NewRes>,
        NewPrec: prec::State,
    {
        let result;

        self.rb.csr.write(|w| {
            use func::reg_count::arg::State as _;
            use func::reg_count::res::State as _;
            use func::scale::State as _;
            result = (
                NewArg::set(w.argsize()),
                NewRes::set(w.ressize()),
                NewPrec::set(w.precision()),
                NewFunc::set(w.func()),
                NewFunc::NArgs::set(w.nargs()),
                NewFunc::NRes::set(w.nres()),
                NewFunc::Scale::set(w.scale()),
           );

            w
        });

        result
    }
```

But rust does not accept the deferred initialization of `result` in this context... for some reason. And the closure passed to `write` *must* return `&mut W`. So I think I'm out of options :(

Other than that I think everything is fine. I can continue to torment myself *outside* of this PR :)

usbalbin (approved):

usbalbin:
Sorry for late answer

> [..] But rust does not accept the deferred initialization of `result` in this context... for some reason. And the closure passed to `write` _must_ return `&mut W`. So I think I'm out of options :( [..]

I guess rust does not know that `write` will actually call the closure so it can not guarantee that `result` will be initialized. I suppose you could (with some care) use something like [`MaybeUninit`](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html) since we do know that `write` will call that closure.

Still, requires unsafe so not ideal... But I believe it would still enforce type safety, right?

AdinAck:
> But I believe it would still enforce type safety, right?

It would certainly *work* but it wouldn't enforce anything. One could easily still comment out any call to `set` and it would compile and exhibit undefined behavior.

I actually ended up getting this working by forking svd2rust and changing `write` and `modify` to support arbitrary return types. Then I rebuilt the PAC and went through all of the HAL and converted everything over to the new style (getters).

The result was perfection :)

```rust
pub fn freeze<NewArg, NewRes, NewPrec, NewOp>(self) -> Cordic<NewArg, NewRes, NewPrec, NewOp>
where
    NewArg: types::arg::State,
    NewRes: types::res::State,
    NewPrec: prec::State,
    NewOp: op::sealed::Feature,
    NewOp::NArgs<NewArg>: reg_count::arg::State,
    NewOp::NRes<NewRes>: reg_count::res::State,
    NewOp::Scale: scale::State,
    NewOp::Func: func::State,
{
    use func::State as _;
    use reg_count::arg::State as _;
    use reg_count::res::State as _;
    use scale::State as _;

    let config = self.rb.csr().modify(|_, w| Config {
        arg: NewArg::set(w.argsize()),
        res: NewRes::set(w.ressize()),
        nargs: NewOp::NArgs::set(w.nargs()),
        nres: NewOp::NRes::set(w.nres()),
        scale: NewOp::Scale::set(w.scale()),
        prec: NewPrec::set(w.precision()),
        func: NewOp::Func::set(w.func()),
    });

    Cordic {
        rb: self.rb,
        config,
    }
}
```

usbalbin:
That is really nice!

What should we do in the mean time? :)

AdinAck:
It works as is, there is some kind of bug somewhere because the resulting binary is smaller than doing the register writes manually. My finalized type-enforced version has no such bug. I am no longer working on the version in this PR, so it could be merged for now, and eventually when the PAC and svd2rust are updated, I can re-PR the final version.

usbalbin:
Not sure I understand, but do you know if the `MaybeUninit` thing solves that?

AdinAck:
> Not sure I understand

I set up some tests where I configure the CORDIC with the PAC manually, and then do the same with the abstraction, and compare the resulting binaries. The version in this PR resulted in a different binary, so clearly it is not doing the exact same thing. With the version in my fork, [this does not happen](https://adinack.dev/blog/better-hals-first-look/#performance). I imagine there is a very subtle error somewhere that is not being caught by type-enforcement because that system is not in place.

> do you know if the MaybeUninit thing solves that

I believe it *could*, I'm not sure, but I don't particularly want to pursue that avenue myself as I don't like the hacky and incomplete nature of it, sorry.

usbalbin:
> > Not sure I understand
> 
> I set up some tests where I configure the CORDIC with the PAC manually, and then do the same with the abstraction, and compare the resulting binaries. The version in this PR resulted in a different binary, so clearly it is not doing the exact same thing. With the version in my fork, [this does not happen](https://adinack.dev/blog/better-hals-first-look/#performance). I imagine there is a very subtle error somewhere that is not being caught by type-enforcement because that system is not in place.

Thanks for the read! :)

> > do you know if the MaybeUninit thing solves that
> 
> I believe it _could_, I'm not sure, but I don't particularly want to pursue that avenue myself as I don't like the hacky and incomplete nature of it, sorry.

That's fair, no worries :)

I will merge this shortly. Again, thanks a lot!

AdinAck:
Haha, of course!

PR #133, opened 2024-08-24
Remove unneeded cfgs from examples causing errors on non G474-devices

usbalbin:


usbalbin:
Thanks @AdinAck for finding the problem

PR #50, opened 2022-04-18
USB device support and usb_serial example

kevswims:
Draft PR to go in after the separate https://github.com/stm32-rs/stm32g4xx-hal/pull/49 to ease review.

This PR adds the following features:
- Updates dependencies
- Updates vscode settings for rust analyzer
- Adds better defmt support along with using probe-run instead of openocd
- Adds USB device support along with a working usb_serial example

This has been tested on a custom STM32F473 board. If desired I can split this out into separate PRs for the defmt/probe-run and the USB support.

PR #135, opened 2024-08-30
USB peripheral support

liamkinne:
Adds the required implementation to support USB device with `usb-device` and `stm32-usbd`.

### DIFF
--- a/src/usb.rs
+++ b/src/usb.rs
@@ -1,0 +1,69 @@
+//! USB peripheral.
+//!
+//! Provides the required implementation for use of the [`stm32-usbd`] crate.
+
+pub use stm32_usbd::UsbBus;
+
+use crate::gpio;
+use crate::gpio::gpioa::{PA11, PA12};
+use crate::rcc::{Enable, Reset};
+use crate::stm32::{RCC, USB};
+use core::fmt;
+use stm32_usbd::UsbPeripheral;
+
+/// Trait implemented by all pins that can be the "D-" pin for the USB peripheral
+pub trait DmPin: crate::Sealed {}
+
+/// Trait implemented by all pins that can be the "D+" pin for the USB peripheral
+pub trait DpPin: crate::Sealed {}
+
+impl DmPin for PA11<gpio::Alternate<{ gpio::AF14 }>> {}
+impl DpPin for PA12<gpio::Alternate<{ gpio::AF14 }>> {}
+
+pub struct Peripheral<Dm: DmPin, Dp: DpPin> {
+    /// USB register block
+    pub usb: USB,
+    /// Data negative pin
+    pub pin_dm: Dm,
+    /// Data positive pin
+    pub pin_dp: Dp,
+}
+
+impl<Dm, Dp> fmt::Debug for Peripheral<Dm, Dp>
+where
+    Dm: DmPin + fmt::Debug,
+    Dp: DpPin + fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Peripheral")
+            .field("usb", &"USB")
+            .field("pin_dm", &self.pin_dm)
+            .field("pin_dp", &self.pin_dp)
+            .finish()
+    }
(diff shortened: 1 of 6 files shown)


liamkinne:
@usbalbin I have mostly ripped this implementation from the F3 repo. However I'm not sure if it is not working correctly or if I have a hardware issue since the G4 board I'm using doesn't feature a USB connector.

If you have the ability to test using the example that would be great. Otherwise wait until I can get my hands on the correct hardware before merging.

src/usb.rs
```diff
@@ -0,0 +1,69 @@
...
+// SAFETY: The peripheral has the same regiter blockout as the STM32 USBFS
+unsafe impl<Dm: DmPin + Send, Dp: DpPin + Send> UsbPeripheral for Peripheral<Dm, Dp> {
+    const REGISTERS: *const () = USB::ptr().cast::<()>();
+    const DP_PULL_UP_FEATURE: bool = true;
+    const EP_MEMORY: *const () = 0x4000_6000 as _;
+    const EP_MEMORY_SIZE: usize = 1024;
+    const EP_MEMORY_ACCESS_2X16: bool = true;
+
+    fn enable() {
+        critical_section::with(|_| unsafe {
```
usbalbin, line 59: In other peripherals we pass a mutable reference to `Rcc` to ensure exclusive access. Would that work here?

usbalbin:
Note there is also #50 . @kevswims what is the status and/or plans of #50?

usbalbin:
> @usbalbin I have mostly ripped this implementation from the F3 repo. However I'm not sure if it is not working correctly or if I have a hardware issue since the G4 board I'm using doesn't feature a USB connector.
> 
> If you have the ability to test using the example that would be great. Otherwise wait until I can get my hands on the correct hardware before merging.

I do not have access to such hardware right now. Might in several months, maybe, but not now.

BTW, thanks!

src/usb.rs
usbalbin: Oh wait, it is implementing a trait from outside this crate...

liamkinne:
@usbalbin didn't see #50 but I will have a look at it an see if there's anything to bring across.

I'll go ahead and order the expansion board to give my dev board USB. If that doesn't fix it, my best guess is there's a clocking issue causing the host to reject the connection.

usbalbin:
> @usbalbin didn't see #50 but I will have a look at it an see if there's anything to bring across.
> 
> I'll go ahead and order the expansion board to give my dev board USB. If that doesn't fix it, my best guess is there's a clocking issue causing the host to reject the connection.

Awesome!

Correct me if I am wrong, have not looked into USB, but I believe I read about some sort of clock recovery system in #50 for running without a crystal in case that helps.

kevswims:
> Note there is also #50 . @kevswims what is the status and/or plans of #50?

I have switched my project over to use embassy which is why I haven't done anything more with that PR. As far as I remember with a bit of cleanup it should be good to go in if someone wants to use the work from it. I'm happy to answer questions or assist with that if someone wants to take it over.

liamkinne:
@usbalbin with the correct hardware, all is working now.

src/usb.rs
usbalbin: What is the different between `critical_section::with` (requiring a new dependency) and `cortex_m::interrupt::free` as used in [src/serial/usart.rs](https://github.com/stm32-rs/stm32g4xx-hal/blob/main/src/serial/usart.rs#L194)?

usbalbin: If there would be no downside, I would prefer to avoid the extra dependency

liamkinne:
@usbalbin this is ready to merge. The failing CI is unrelated.

usbalbin:
Great!

Perhaps not a huge deal, but would you mind having a look at my [earlier comment](https://github.com/stm32-rs/stm32g4xx-hal/pull/135#discussion_r1755370264) :)

liamkinne:
@usbalbin sorry didn't see that. Fixed now.

usbalbin:
Thanks a lot for this!

PR #137, opened 2024-09-01
rcc: Correct apb1_freq and apb2_freq calculations.

projectgus:
As per the Reference Manual Clock Tree (figure 17), these frequencies are divided from `ahb_freq` (HCLK) not `sys_clk` (SYSCLK).

Segment of Figure 17

Thanks for all your work maintaining this crate!

### DIFF
--- a/src/rcc/mod.rs
+++ b/src/rcc/mod.rs
@@ -109,18 +109,18 @@ impl Rcc {
             _ => (sys_freq, 0b0000),
         };
         let (apb1_freq, apb1_psc_bits) = match rcc_cfg.apb1_psc {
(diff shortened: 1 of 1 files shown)


usbalbin:
Wow, nice find! Thank you :)

usbalbin:
From what I can tell, the affected peripherals are:
* All TIMx, LPTIMx, and HRTIM1 timers
* All UARTs and I2Cx depending on clock source

projectgus:
@usbalbin Thanks for the quick review and merge!

FDCAN is affected as well (that's how I ran into it.) FDCAN can select its clock from multiple sources, but currently the HAL sets it to PCLK1.

PR #138, opened 2024-09-08
Allow to "freeze" gpio type state

usbalbin:
### 1. Changing pin state
Today a gpio pin can be turned into another mode using for example `let pin = pin.into_push_pull_output();` or `let pin = pin.into_alternate()`. This can also be done multiple times for example:

```rust
let pin = pin.into_floating_input(); // Configure as digital input
//use as input...

let mut pin = pin.into_push_pull_output(); // Reconfigure the same pin, now as an output
pin.set_low();

let pin = pin.into_analog(); // Reconfigure as analog input
// perform AD-reading...
```

This is of course very convenient.

### 2. Same pin, Multiple peripherals

Some analog peripherals like for example the comparators need their input pins to be set in analog mode. Since the pin is set to analog mode it should also be possible to perform AD-readings on it at the same time as the comparator is connected.

This may for example be useful for having near instant over current protection while still allowing regular AD-readings all on the same pin.

### 3. Peripherals

When setting up a peripheral such as the comparator, the input pins are typically provided by value, thus consuming the pin. This to ensure that no one else will change the mode of the pin.

### The problem

Currently `1` and `3` works, but not `2`. Changing peripherals to take a pin by reference and store the reference would prove quite cumbersome since the pin would have to live as long as the peripheral and not be moved etc. 

Only having having the init function take a reference and then not storing it would not prevent the user from then changing pin mode after peripheral init.

### Proposed solution
Add type parameter `F` to the pin types which may be one of the following types:

```diff
- pub struct $PXi<MODE> {
+ pub struct $PXi<MODE, F = IsNotFrozen> {
     _mode: PhantomData<MODE>,
+    _frozen_state: PhantomData<F>,
 }

```rust
/// Type state for a gpio pin frozen to its current mode
/// 
/// The into_alternate/analog/{x}_input/output can not be called on a pin with this state.
/// The pin is thus guaranteed to stay as this type as long as the program is running.
pub struct IsFrozen;

/// See [IsFrozen]
pub struct IsNotFrozen;
```
The type parameter default to `IsNotFrozen` so as to keep backwards compatibility and `1` from above works as before. However once the user knows they will want the pin to stay in a particular state for the rest of the program, `let pin = pin.freeze();` may be called. Which returns the same type but with `IsFrozen` this disabling the `into_alternate/analog/{x}_input/output` methods.

This then allows peripherals init to take the pin by reference (assuming `F=IsFrozen`) and not having to hold on to it. (only done for some of the peripherals so far)

### DIFF
--- a/examples/comp.rs
+++ b/examples/comp.rs
@@ -46,7 +46,7 @@ fn main() -> ! {
             .output_inverted(),
         &rcc.clocks,
     );
-    let led2 = gpioa.pa12.into_push_pull_output();
+    let led2 = gpioa.pa12.into_alternate();
     // Configure PA12 to the comparator's alternate function so it gets
     // changed directly by the comparator.
     comp2.output_pin(led2);
--- a/examples/comp_w_dac.rs
+++ b/examples/comp_w_dac.rs
@@ -30,7 +30,7 @@ fn main() -> ! {
     // Set up DAC to output to pa4 and to internal signal Dac1IntSig1
     // which just so happens is compatible with comp1
     let dac1ch1 = dp.DAC1.constrain((gpioa.pa4, Dac1IntSig1), &mut rcc);
-    let mut dac = dac1ch1.calibrate_buffer(&mut delay).enable();
+    let mut dac = dac1ch1.calibrate_buffer(&mut delay).enable().freeze();
 
     let (comp1, _comp2, ..) = dp.COMP.split(&mut rcc);
     let pa1 = gpioa.pa1.into_analog();
--- a/examples/comp_w_dac.rs
+++ b/examples/comp_w_dac.rs
@@ -43,7 +43,7 @@ fn main() -> ! {
         &rcc.clocks,
     );
 
-    let led2 = gpioa.pa0.into_push_pull_output();
+    let led2 = gpioa.pa0.into_alternate();
     // Configure PA12 to the comparator's alternate function so it gets
     // changed directly by the comparator.
     comp.output_pin(led2);
--- a/examples/opamp.rs
+++ b/examples/opamp.rs
@@ -44,7 +44,7 @@ fn main() -> ! {
 
     // Configure opamp1 with pa1 as non-inverting input and set gain to x2
     let _opamp1 = opamp1.pga(&pa1, pa2, Gain::Gain2);
-    
+
     // Set up opamp2 in follower mode with output routed to internal ADC
     let opamp2 = opamp2.follower(&pa7, InternalOutput);
 
--- a/examples/opamp.rs
+++ b/examples/opamp.rs
@@ -31,35 +31,25 @@ fn main() -> ! {
     let gpioa = dp.GPIOA.split(&mut rcc);
     let gpiob = dp.GPIOB.split(&mut rcc);
 
+    let pa1 = gpioa.pa1.into_analog();
+    let pa2 = gpioa.pa2.into_analog();
+    let pa7 = gpioa.pa7.into_analog();
+
+    let pb0 = gpiob.pb0.into_analog();
+    let pb1 = gpiob.pb1.into_analog();
+    let pb2 = gpiob.pb2.into_analog();
+
     // setup opamps
     let (opamp1, opamp2, opamp3, ..) = dp.OPAMP.split(&mut rcc);
 
-    // Set up opamp1 and opamp2 in follower mode
-    let opamp1 = opamp1.follower(gpioa.pa1, gpioa.pa2);
-    let opamp2 = opamp2.follower(gpioa.pa7, InternalOutput);
-
-    // Set up opamp1 and opamp2 in open loop mode
-    let opamp3 = opamp3.open_loop(gpiob.pb0, gpiob.pb2, gpiob.pb1);
-
-    // disable opamps
-    let (opamp1, pa1, pa2) = opamp1.disable();
-    let (opamp2, pa7) = opamp2.disable();
-
-    let (_opamp3, _pb0, _pb2, _pb1) = opamp3.disable();
-
     // Configure opamp1 with pa1 as non-inverting input and set gain to x2
-    let _opamp1 = opamp1.pga(
-        pa1,
-        pa2, // Route output to pin pa2
-        Gain::Gain2,
-    );
+    let _opamp1 = opamp1.pga(&pa1, pa2, Gain::Gain2);
+    
+    // Set up opamp2 in follower mode with output routed to internal ADC
+    let opamp2 = opamp2.follower(&pa7, InternalOutput);
 
-    // Configure op with pa7 as non-inverting input and set gain to x4
-    let opamp2 = opamp2.pga(
-        pa7,
-        InternalOutput, // Do not route output to any external pin, use internal AD instead
-        Gain::Gain4,
-    );
+    // Set up opamp3 in open loop mode
+    let _opamp3 = opamp3.open_loop(&pb0, pb2, pb1);
 
     // Lock opamp2. After the opamp is locked its registers cannot be written
     // until the device is reset (even if using unsafe register accesses).
@@ -78,15 +68,13 @@ fn main() -> ! {
         );
 
         let millivolts = adc.sample_to_millivolts(sample);
-        info!("opamp2 thus 4x pa7: {}mV", millivolts);
+        info!("opamp2 thus pa7: {}mV", millivolts);
 
         delay.delay_ms(100);
     }
 
     #[allow(unreachable_code)]
     {
-        let (_opamp1, _pin) = _opamp1.disable();
-
         loop {
             delay.delay_ms(100);
         }
--- a/src/comparator.rs
+++ b/src/comparator.rs
@@ -10,7 +10,6 @@ use core::marker::PhantomData;
 
 use crate::dac;
 use crate::exti::{Event as ExtiEvent, ExtiExt};
-use crate::gpio::{FrozenPin, IsFrozen};
 use crate::gpio::{
     gpioa::{PA0, PA1, PA11, PA12, PA2, PA3, PA4, PA5, PA6, PA7},
     gpiob::{PB0, PB1, PB14, PB15, PB2, PB6, PB7, PB8, PB9},
@@ -18,6 +17,7 @@ use crate::gpio::{
     gpiof::PF4,
     Alternate, AlternateOD, Analog, SignalEdge, AF2, AF3, AF8,
 };
+use crate::gpio::{FrozenPin, IsFrozen};
 
 #[cfg(any(
     feature = "stm32g473",
@@ -143,9 +143,9 @@ pub enum Hysteresis {
 /// Comparator positive input
 pub trait PositiveInput<C>: crate::Sealed {
     /// Setup comaprator to use this as its positive input
-    /// 
+    ///
     /// # Safety
-    /// 
+    ///
     /// Internal use only
     unsafe fn setup(&self, comp: &mut C);
 }
@@ -165,7 +165,7 @@ pub trait NegativeInput<C>: crate::Sealed {
     /// Setup comaprator to use this as its negative input
     ///
     /// # Safety
-    /// 
+    ///
     /// Internal use only
     unsafe fn setup(&self, comp: &mut C);
 }
@@ -436,7 +436,7 @@ macro_rules! impl_comparator {
             ) -> Comparator<$COMP, Disabled>
             where
                 PP: FrozenPin<P> + PositiveInput<$COMP>,
-                NP: FrozenPin<N> + NegativeInput<$COMP>
+                NP: FrozenPin<N> + NegativeInput<$COMP>,
             {
                 unsafe {
                     positive_input.borrow().setup(&mut self);
@@ -474,7 +474,7 @@ macro_rules! impl_comparator {
             ) -> Self
             where
                 PP: FrozenPin<P> + PositiveInput<$COMP>,
-                NP: FrozenPin<N> + NegativeInput<$COMP>
+                NP: FrozenPin<N> + NegativeInput<$COMP>,
             {
                 comp.comparator(positive_input, negative_input, config, clocks)
             }
--- a/src/comparator.rs
+++ b/src/comparator.rs
@@ -17,7 +17,7 @@ use crate::gpio::{
     gpiof::PF4,
     Alternate, AlternateOD, Analog, SignalEdge, AF2, AF3, AF8,
 };
-use crate::gpio::{FrozenPin, IsFrozen};
+use crate::gpio::{Frozen, IsFrozen};
 
 #[cfg(any(
     feature = "stm32g473",
@@ -281,7 +281,7 @@ pub enum RefintInput {
     VRefint = 0b011,
 }
 
-impl FrozenPin<RefintInput> for RefintInput {}
+impl Frozen<RefintInput> for RefintInput {}
 impl crate::Sealed for RefintInput {}
 
 macro_rules! refint_input {
@@ -420,8 +420,8 @@ pub trait ComparatorExt<COMP> {
         clocks: &Clocks,
     ) -> Comparator<COMP, Disabled>
     where
-        PP: FrozenPin<P> + PositiveInput<COMP>,
-        NP: FrozenPin<N> + NegativeInput<COMP>;
+        PP: Frozen<P> + PositiveInput<COMP>,
+        NP: Frozen<N> + NegativeInput<COMP>;
 }
 
 macro_rules! impl_comparator {
@@ -435,8 +435,8 @@ macro_rules! impl_comparator {
                 clocks: &Clocks,
             ) -> Comparator<$COMP, Disabled>
             where
-                PP: FrozenPin<P> + PositiveInput<$COMP>,
-                NP: FrozenPin<N> + NegativeInput<$COMP>,
+                PP: Frozen<P> + PositiveInput<$COMP>,
+                NP: Frozen<N> + NegativeInput<$COMP>,
             {
                 unsafe {
                     positive_input.borrow().setup(&mut self);
@@ -473,8 +473,8 @@ macro_rules! impl_comparator {
                 clocks: &Clocks,
             ) -> Self
             where
-                PP: FrozenPin<P> + PositiveInput<$COMP>,
-                NP: FrozenPin<N> + NegativeInput<$COMP>,
+                PP: Frozen<P> + PositiveInput<$COMP>,
+                NP: Frozen<N> + NegativeInput<$COMP>,
             {
                 comp.comparator(positive_input, negative_input, config, clocks)
             }
--- a/src/comparator.rs
+++ b/src/comparator.rs
@@ -14,7 +14,7 @@ use crate::gpio::{
     gpiob::{PB0, PB1, PB14, PB15, PB2, PB6, PB7, PB8, PB9},
     gpioc::PC2,
     gpiof::PF4,
-    Analog, OpenDrain, Output, PushPull, SignalEdge, AF2, AF3, AF8,
+    Analog, Alternate, AlternateOD, SignalEdge, AF2, AF3, AF8,
 };
 
 #[cfg(any(
@@ -486,9 +486,7 @@ macro_rules! impl_comparator {
             /// Configures a GPIO pin to output the signal of the comparator
             ///
             /// Multiple GPIO pins may be configured as the output simultaneously.
-            pub fn output_pin<P: OutputPin<$COMP>>(&self, pin: P) {
-                pin.setup();
-            }
+            pub fn output_pin<P: OutputPin<$COMP>>(&self, _pin: P) {}
         }
     };
 }
@@ -587,22 +585,16 @@ impl ComparatorSplit for COMP {
     }
 }
 
-pub trait OutputPin<COMP> {
-    fn setup(self);
-}
+pub trait OutputPin<COMP> {}
 
 #[allow(unused_macros)] // TODO: add support for more devices
 macro_rules! output_pin {
-    ($COMP:ident, $pin:ident, $AF:ident, $mode_t:ident, $into:ident) => {
-        impl OutputPin<$COMP> for $pin<Output<$mode_t>> {
-            fn setup(self) {
-                self.$into::<$AF>();
-            }
-        }
+    ($COMP:ident, $pin:ident, $AF:ident, $mode_t:ident) => {
+        impl OutputPin<$COMP> for $pin<$mode_t<$AF>> {}
     };
     ($($COMP:ident: $pin:ident, $AF:ident,)+) => {$(
-        output_pin!($COMP, $pin, $AF, PushPull, into_alternate);
-        output_pin!($COMP, $pin, $AF, OpenDrain, into_alternate_open_drain);
+        output_pin!($COMP, $pin, $AF, Alternate);
+        output_pin!($COMP, $pin, $AF, AlternateOD);
     )+};
 }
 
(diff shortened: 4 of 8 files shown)


usbalbin:
@AdinAck do you have any input on this?

AdinAck:
I have also been thinking about this!

It's really quite unfortunate that references to ZSTs are not ZSTs themselves, otherwise I believe holding references would have been the correct solution, the lifetime troubles could certainly be worked out.

However, since we cannot do that without introducing significant overhead, this is a clever solution. I can't tell if it's *complete* though...

I believe that the notion that a peripheral shall be **owned** by one thing and one thing only is an invariance that should not be broken. Seeing as the goal of the HAL is to maintain a certain parity with the hardware, what if instead we change the HAL pin structs a bit to reflect the hardware? Clearly comparators have an additional special access to select pins.

So the pins could have channels:

```rust
struct PXi<Mode> {
    ch1: PXi1,
    ch2: PXi2,
```

That way, it can be used like so:

```rust
let pin = gpiox.pi.into_input_pull_up().split(); // `split()` will only be implemented for inputs

// i don't know either of these peripheral interfaces
let adc = p.ADC.constrain(pin.ch1);
let comp = p.COMP.constrain(pin.ch2);
```

Conversion methods (`into_x`) will not be available on the channels, so the pin is effectively frozen. The inverse of `split()` can be provided as `recombine()` (this is an idea I've been playing around with as a standard for all split peripherals):

```rust
let pin = pin.common.recombine((adc.reset_and_release().1, comp.reset_and_release().1)).into_push_pull_output();
```

> `pin.common` is inspired by the PWM interface.

---

I'm still thinking about all this, so it's all in flux. I think this better adheres to the design patterns already present in the HAL, but maybe it's too complicated, let me know what you think. I am not against your proposal, it's certainly a step in the right direction nonetheless.

AdinAck:
Example of how `IsFrozen` applies to other peripherals:

```rust
let gpiox = p.GPIOX.split();
let px1 = gpiox.px1.into_analog().freeze();
let px2 = gpiox.px2.into_floating_input().freeze();

let (dacx_ch1, ..) = p.DACX.constrain();
let dacx_ch1 = dacx_ch1.freeze(); // dac channel must be frozen for use with comp

let exti = p.EXTI.split();
let (comp1, ..) = p.COMP.split();
let comp1 = comp1.constrain(&dacx_ch1, &px1); // requires both sources are frozen
let comp1 = comp1.freeze(); // now even the comp itself is frozen for use with EXTI

exti.ch2.constrain(&px2).enable(); // requires `px2` is frozen
exti.ch23.constrain(&comp1).enable(); // requires `comp1` is frozen
```

This example also demonstrates an improved and more consistent EXTI interface.

AdinAck:
There should also be a `Frozen` trait which can be added as a requirement to source traits.

Like for comparators:

```rust
pub trait PositiveInput<C>: Frozen {
    fn setup(&self, comp: &C);
}
```

usbalbin:
> So the pins could have channels: [...]

What about something like:
```rust
fn split<const N: usize>(self) -> [Channel<N>; N] {
    todo!()
}
fn recombine(..., channels: [Channel<N>; N]) -> ... {
    todo!()
}
```
The user is free to split into however many channels they want. The channels then remember how many siblings they are so that we can require them all when recombining

AdinAck:
> What about something like...

A fantastic idea! I will need time to think about this :)

AdinAck:
Ok here's what I've come up with:

What we're dealing with here is a new way that peripherals interact with each other. In this case, EXTI, COMP, and ADC are *observing* GPIO. To facilitate this, consider:

```rust
struct Observed<P, const OBSERVERS: usize> {
    p: P,
}
```
> A struct to hold peripherals which are to be observed.

```rust
struct ObservationToken<P> {
    _p: PhantomData<P>,
}
```
> A struct to represent a registered observation of a peripheral.

```rust
trait Observable {
    fn observe<const N: usize>(self) -> (Observed<Self, N>, [ObservationToken<Self>; N]) {
        (
            Observed { p: self },
            core::array::from_fn(|| ObservationToken { _p: PhantomData }), // this may be introduce overhead, will have to investigate
        )
    }
}
```
> A trait providing an interface to make peripherals observed.

Usage:

```rust
let (pin, [ob_exti, ob_comp, ob_adc]) = gpiox.pxi.into_analog().observe();
// pin type is `Observed<PXi<Analog>, 3>`
// ob_* type is `ObservationToken<PXi<Analog>>`

exti.source(ob_exti).enable();
comp.source(dac_ch1, ob_comp);
adc.constrain(Sequence { .., third: ob_adc }); // very pseudo

let (_, ob_exti) = exti.reset_and_release();
let (_, ob_comp) = comp.reset_and_release();
let (_, ob_adc) = adc.reset_and_release();

let pin = pin.release([ob_exti, ob_comp, ob_adc]);

pin.get_level(); // we can use the pin in software again
```

The key to my reasoning for this design is these peripheral interfaces only need to "lock" the observed peripheral configuration, and **do not** invoke any of its methods because the interaction is purely via hardware. Only software interfaces should borrow peripherals by reference.

So HAL interfaces will look like:

```rust
fn constrain<INP, INM>(rb: COMP1, inp: ObservationToken<INP>, inm: ObservationToken<INM>)
where
    INP: PositiveInput,
    INM: NegativeInput,
{ .. }
```

Since the HAL never uses the software methods (really it *shouldn't* anyways) the usage of `ObservationToken` explicitly shows the hardware relationship. (The name may be poor but of course, that can be worked out).

This whole design philosophy relies on the invariant fact that up to one instance of any peripheral type exists.

---

Still more thinking to be done, some holes to patch, please let me know your thoughts.

AdinAck:
I made a quick test in [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:2,endLineNumber:36,positionColumn:2,positionLineNumber:36,selectionStartColumn:2,selectionStartLineNumber:36,startColumn:2,startLineNumber:36),source:'use+std::marker::PhantomData%3B%0A%0A%23%5Bderive(Debug)%5D%0Astruct+Analog%3B%0A%0A%23%5Bderive(Debug)%5D%0Astruct+PXi%3CMode%3E+%7B%0A++++_mode:+PhantomData%3CMode%3E%0A%7D%0A%0A%23%5Bderive(Debug)%5D%0Astruct+Observed%3CP,+const+N:+usize%3E+%7B%0A++++p:+P,%0A%7D%0A%0A%23%5Bderive(Debug)%5D%0Astruct+ObservationToken%3CP%3E+%7B%0A++++_p:+PhantomData%3CP%3E,%0A%7D%0A%0Atrait+Observable:+Sized+%7B%0A++++fn+observe%3Cconst+N:+usize%3E(self)+-%3E+(Observed%3CSelf,+N%3E,+%5BObservationToken%3CSelf%3E%3B+N%5D)+%7B%0A++++++++(%0A++++++++++++Observed+%7B+p:+self+%7D,%0A++++++++++++core::array::from_fn(%7C_%7C+ObservationToken+%7B+_p:+PhantomData+%7D),%0A++++++++)%0A++++%7D%0A%7D%0A%0Aimpl%3CMode%3E+Observable+for+PXi%3CMode%3E+%7B+%7D%0A%0Afn+main()+%7B%0A++++let+(pin,+%5Bob0,+ob1%5D)+%3D+PXi+%7B+_mode:+PhantomData::%3CAnalog%3E+%7D.observe()%3B%0A%0A++++println!!(%22%7B:%3F%7D,+%7B:%3F%7D,+%7B:%3F%7D%22,+pin,+ob0,+ob1)%3B%0A%7D'),l:'5',n:'1',o:'Rust+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:r1810,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'0',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-C++++++++++++++++opt-level%3D3',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+rustc+1.81.0+(Editor+%231)',t:'0')),k:50,l:'4',m:80.8411214953271,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+7.3',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.81.0+(Compiler+%231)',t:'0')),header:(),l:'4',m:19.158878504672895,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',m:100,n:'0',o:'',t:'0')),version:4) and [rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3488342e524a7054607fbe9c798b2ada) and I think the usage of `core::array::from_fn` does not add overhead.

usbalbin:
> ```rust
> let (pin, [ob_exti, ob_comp, ob_adc]) = gpiox.pxi.into_analog().observe();
> // pin type is `Observed<PXi<Analog>, 3>`
> // ob_* type is `ObservationToken<PXi<Analog>>`
>
> exti.source(ob_exti).enable();
> comp.source(dac_ch1, ob_comp);
> adc.constrain(Sequence { .., third: ob_adc }); // very pseudo
>
> ```

Just trying to wrap my head around this :) Is `pin` here more or less the same as a frozen pin from my first proposal? So all `pin.into_{x}` disabled but the pin is otherwise the same as normal with `pin.is_high` etc available and adc readings possible, assuming the pin is in analog mode?

Or is it more locked down than that? If so, then what use do we even have for pin besides keeping track of the number of observers? Is this better than the observers themselves keeping track of how many siblings they are? Thus removing the need for the `Observed` it self.

> ```rust
> let (_, ob_exti) = exti.reset_and_release();
> let (_, ob_comp) = comp.reset_and_release();
> let (_, ob_adc) = adc.reset_and_release();
>
> let pin = pin.release([ob_exti, ob_comp, ob_adc]);
>
> pin.get_level(); // we can use the pin in software again
> ```

What does "use the pin in software" mean? Was `pin.is_high()` disabled until now? How about software triggered adc readings, is that "software"? Or am I supposed to use `ObservationToken` even for software triggered adc readings?

Lots of questions... Just trying to wrap my head around what consequences something like this would result in. :)

AdinAck:
> Is `pin` here more or less the same as a frozen pin from my first proposal? So all `pin.into_{x}` disabled but the pin is otherwise the same as normal with `pin.is_high` etc available and adc readings possible, assuming the pin is in analog mode?

Kind of, since it's of type `Observed<..>`, we should probably use `AsRef` for convenience:

```rust
impl<P, const N: usize> AsRef<P> for Observed<P, N> {
    fn as_ref(&self) -> &P {
        self.p
    }
}
```

This is great because methods like `is_high` are available and all mutating functions `set_high` are *not* available (why should they be?).

> How about software triggered adc readings, is that "software"? Or am I supposed to use `ObservationToken` even for software triggered adc readings?

A software trigger of the ADC *would not* be the same as calling `set_high` for example on a pin. An ADC conversion only mutates the ADC peripheral state, the ADC is just *observing* that pin. The HAL interface for the ADC being used on certain pins *should* only have an `ObservationToken` to the pin since the measurement does not require calling any pin methods, `is_high`, `set_high`, etc.

In conclusion, if I understand your questions correctly:
- should pins be usable by software while observed? **yes, immutably**
- should HAL interfaces for peripherals which use other peripherals but *don't* invoke their methods have access to those methods? **no, they should have an `ObservationToken`**

It *could* be argued that one *should* be able to mutate a peripheral while it is observed, this would be useful for testing, not production. I feel that additional interface is misleading and extraneous, but I could understand the desire for its presence.

Also, I share your concern for introducing this interface, it's a rather sizable change, I agree it is important to make sure this idea is absolutely the correct decision, so I appreciate the questions! :)

AdinAck:
You know, adding an impl for `AsMut` is dead easy, might as well add it as well!

usbalbin:
> [...]I agree it is important to make sure this idea is absolutely the correct decision, so I appreciate the questions! :)

Thanks a lot for your input on this :)

> Also, I share your concern for introducing this interface, it's a rather sizable change,[...]

Correct me if I am wrong, but a lot (perhaps not all but still, a lot) of this should be possible to do in a backwards compatible way by still allowing the owned(not splitted) peripheral/pin in places where a Observer or Observed are expected, right?

AdinAck:
Hm, I'm not sure! Only one way to find out, I'll test it out and see what happens :)

AdinAck:
Ah, yes. We can implement `Into<ObservationToken<Self>` for all peripherals, so old code will still work.

usbalbin:
See #157 for an attempt at implementing that idea

usbalbin:
Some random thoughts:

Side note, it would be quite interesting to see what could be done with a system like this for the rcc and its consumers. At the very least i suppose that would allow a safe way to change clock frequency once tokens from all consumer are collected without risk of forgetting any peripheral that might not like the clock settings to change.

However doing something like that well is likely not trivial, perhaps requiring something more fine grained? Would if the RTC is clocked by the LSE then changing other parts of the clock tree should not affect the RTC etc.

Side note to the Side note. Thera are probably a lot of interesting things that can be done to the rcc system more generally. Lots of weird hardware limitations with the different steps in the pll system having different max frequencies etc. Unless am completely misremembering I believe the max frequency for the ADC is 60MHz but only if you just use one of them, if you use more then I think just north of 50. The hrtim has a *minimum* input frequency of 100MHz. There are probably lots of more weird things that would be interesting to see if it would be possible to check for at compile time. Then there are different limitations depending on input voltage etc.

What if the user could use a const fn to generate a clock config that was checked at compile time, that config was then applied at runtime. Perhaps one could even have multiple configs in different constants and it would be completely safe to dynamicly change between them(assuming it would be possible to solve both things above).

AdinAck:
> Side note, it would be quite interesting to see what could be done with a system like this for the rcc and its consumers. At the very least i suppose that would allow a safe way to change clock frequency once tokens from all consumer are collected without risk of forgetting any peripheral that might not like the clock settings to change.

I have been thinking about this for many months and believe that it is the *perfect* opportunity to flex why Rust is so good for embedded.

> However doing something like that well is likely not trivial, perhaps requiring something more fine grained?

I am working on a system that procedurally generates type-enforced peripheral access interfaces and have some examples where this works and stops the developer from forgetting to enable peripheral clocks before usage. [gpio example](https://github.com/AdinAck/proto-hal/blob/5861351fd5ae741c25e8651319eae9eadf88df95/stm32/g4/tests/gpio.rs#L26).

I am very excited to begin reasoning about clock *speeds* in the type system.

> What if the user could use a const fn to generate a clock config that was checked at compile time, that config was then applied at runtime.

This is something I have done in the past in `stm32g0xx-hal`:

```rust
const PLL_CFG: rcc::PllConfig = rcc::RawPllConfig {
    mux: rcc::PLLSrc::HSI, // 16MHz
    m: 1,                  // /1
    n: 16,                 // x16
    r: 4,                  // /4 = 64MHz
    q: Some(2),            // /2 = 128MHz
    p: None,
}
.validate();
```
> This is working code using my fork of `stm32g0xx-hal`.

usbalbin:
> g0xx-hal stuff

Nice! I will take a look at that :)

> gpio example

Very nice! Great to have that level of protection all the way down

How/does this interact with svd-files or does this require every peripheral to be manually specified?

AdinAck:
> How/does this interact with svd-files or does this require every peripheral to be manually specified?

No SVD, they are always riddled with errors from the vendor, which is why we patch them so much. They are hard to read, don't encode any hardware relationships at all, just suck to use.

It is so much easier to just scroll through a datasheet and type in the register/field offsets.

Some brief examples:

```rust
#[block(
    base_addr = 0x4800_0000,
    entitlements = [super::super::rcc::ahb2enr::gpioaen::Enabled],
    auto_increment,
)]
pub mod gpioa {
    #[register(auto_increment)]
    mod moder {
        #[schema(width = 2, auto_increment)]
        mod mode {
            #[state]
            struct Input;
            #[state]
            struct Output;
            #[state]
            struct Alternate;
            #[state]
            struct Analog;
        }

        #[field_array(range = ..13, schema = mode, read, write, reset = Analog)]
        mod modeX {}

        #[field(schema = mode, read, write, reset = Alternate)]
        mod mode13 {}

        #[field(schema = mode, read, write, reset = Alternate)]
        mod mode14 {}

        #[field(schema = mode, read, write, reset = Alternate)]
        mod mode15 {}
    }
```
> The beginning of the description for `GPIOA`.

```rust
#[block(
    base_addr = 0x4002_0c00,
    entitlements = [super::rcc::ahb1enr::cordicen::Enabled],
    auto_increment,
    erase_mod,
)]
mod cordic {
    #[register(auto_increment)]
    mod csr {
        #[field(width = 4, read, write, reset = Cos, auto_increment)]
        mod func {
            #[state(entitlements = [scale::N0])]
            struct Cos;

            #[state(entitlements = [scale::N0])]
            struct Sin;

            #[state(entitlements = [scale::N0])]
            struct ATan2;

            #[state(entitlements = [scale::N0])]
            struct Magnitude;

            #[state]
            struct ATan;

            #[state(entitlements = [scale::N1])]
            struct CosH;

            #[state(entitlements = [scale::N1])]
            struct SinH;

            #[state(entitlements = [scale::N1])]
            struct ATanH;

            #[state(entitlements = [scale::N1, scale::N2, scale::N3, scale::N4])]
            struct Ln;

            #[state(entitlements = [scale::N0, scale::N1, scale::N2])]
            struct Sqrt;
        }
```
> The `func` field of `CORDIC`.

Remember all those crazy types and traits and structures that occupied 1200 lines of code to enforce the CORDIC's hardware invariances? Now they are described in *30* lines of code.

Reading the datasheet is trivial, and this macro statically validates *a lot* in case mistakes are made.

I'm worried my design of this system doesn't scale but I just keep implementing peripherals and adding things as needed. Hopefully eventually it will encapsulate everything.

usbalbin:
> > How/does this interact with svd-files or does this require every peripheral to be manually specified?
> 
> No SVD, they are always riddled with errors from the vendor, which is why we patch them so much. They are hard to read, don't encode any hardware relationships at all, just suck to use.
> 

Yeah, cant say I disagree

> It is so much easier to just scroll through a datasheet and type in the register/field offsets.
> 
> Some brief examples:
> [...]
> Remember all those crazy types and traits and structures that occupied 1200 lines of code to enforce the CORDIC's hardware invariances? Now they are described in _30_ lines of code.

Yes :) Wow!

> Reading the datasheet is trivial, and this macro statically validates _a lot_ in case mistakes are made.

Very nice

> I'm worried my design of this system doesn't scale but I just keep implementing peripherals and adding things as needed. Hopefully eventually it will encapsulate everything.

I think this might be related to https://github.com/stm32-rs/meta/issues/4 .

AdinAck:
> I think this might be related to [stm32-rs/meta#4](https://github.com/stm32-rs/meta/issues/4) .

Very interesting! I'll read this and see if there's anything I can contribute to the conversation, thanks!

usbalbin:
> I'm worried my design of this system doesn't scale but I just keep implementing peripherals and adding things as needed. Hopefully eventually it will encapsulate everything.

Thinking about scaling as in getting something like this to work on many families in the future. As is discussed in the linked issue, there are a lot of devices with a lot of peripherals... Do you think it would be possible and make sense to somehow get this to work with the pacs of today(lots of it which after all the patches of stm32-rs works quite good, yet without all the nice magic type safety things)? Would it even be possible to add the entitlement info and whatnot (sorry have not quite wrapped my head around it) to svd2rust's yaml format so this could be added in peripheral by peripheral while still making use of all the things we already have?

AdinAck:
> Would it even be possible to add the entitlement info and whatnot (sorry have not quite wrapped my head around it) to svd2rust's yaml format

This was my first approach before bailing and making `proto-hal`. It is certainly possible with a great deal of effort.

My motivation for creating `proto-hal` is selfish in nature. I want to *use* it and want it to work exactly how I say. Slowly grossly bandaiding onto the monster that is `svd2rust` sounds no fun! But that may very well be the best way to improve HALs in general. Just not for me.

So far implementing peripherals with the macro has been very easy and quick, so I'm not really worried about redoing work already done in other HALs/PACs, but maybe I'll run into issues soon.

usbalbin:
One perhaps not great idea but at least low effort way would perhaps be if your pac types implemented `From<CorrespondingStm32RsPacType>`? Then the two pacs could be used side by side in case yours was missing a peripheral

AdinAck:
Hm, yeah. Some adaptor between the two would be nice. I'll think on that, thanks!

usbalbin:
This observation-thing feels like it could be useful for other hals and perhaps not only parts from stm. Would it make sense to put src/observation.rs into its own crate to reduce duplication and make it easier to make crates that work with different hals?

AdinAck:
> Would it make sense to put src/observation.rs into its own crate...

Yep! I already have it in `proto-hal` along with other structures I am messing around with for designing HALs in a standardized way.

It's in the [stasis](https://github.com/AdinAck/proto-hal/blob/main/proto-hal/src/stasis.rs) file. I moved on from the "observation" naming scheme because since the beginning we've known that's not the best name. For example if we consider the output level of a pin to *not* contribute to the state of that pin, then an `ObservationToken` would allow you to mutate the pin. It's a bad name. Which is why I moved on to **Entitlements**.

Peripherals are *entitled* to states. Obviously I am using this for far more complicated peripheral interface generation stuff in `macros`, but the rest of `proto-hal` is just for HALs in general and the "observation"/"entitlements" stuff is useful all the same.

AdinAck:
The reason I created `proto-hal` in the first place was because I didn't think the maintainers of `embedded-hal` would be very open to changes, but if you feel otherwise, really that's the ideal place for entitlements.

AdinAck:
I just migrated over most of the important design features to [stasis](https://github.com/AdinAck/proto-hal/blob/main/proto-hal/src/stasis.rs).

usbalbin:
> > Would it make sense to put src/observation.rs into its own crate...
> 
> Yep! I already have it in `proto-hal` along with other structures I am messing around with for designing HALs in a standardized way.
> 
> It's in the [stasis](https://github.com/AdinAck/proto-hal/blob/main/proto-hal/src/stasis.rs) file. I moved on from the "observation" naming scheme because since the beginning we've known that's not the best name. For example if we consider the output level of a pin to _not_ contribute to the state of that pin, then an `ObservationToken` would allow you to mutate the pin. It's a bad name. Which is why I moved on to **Entitlements**.

Oh my bad, sorry I did not get that Entitlements was the same thing as ObservationToken :). Makes sense

> Peripherals are _entitled_ to states. Obviously I am using this for far more complicated peripheral interface generation stuff in `macros`, but the rest of `proto-hal` is just for HALs in general and the "observation"/"entitlements" stuff is useful all the same.
Ah, ok
> I would be open to moving these structures to their _own_ crate all by themselves if you really want. But I feel like it makes sense to keep them in `proto-hal` amongst the other HAL utilities.

Already seems to be quite broken out, no need to break it out further

The rough gist from what I have seen so far(which is not a lot) is:
* proto-hal/proto-hal - Has some types and stuff like the entitlement system and other useful things for gpios etc.
* proto-hal/macros - Has macros to generate a safe pac that uses the entitlement things
* proto-hal/stm32/g4 - Uses the above to specify a safe pac for the G4 family, or something slightly below a hal?

Is that somewhat accurate?

Very interesting stuff :)

AdinAck:
> Is that somewhat accurate?

100% accurate!

> Very interesting stuff :)

I agree! Thanks for all your help!

usbalbin:
Ill close this in favour of #157

PR #139, opened 2024-09-17
Fix `listen` return mode and relax pin constraint

AdinAck:
The `listen` function for GPIO pins had a return type of `P<Input<PushPull>>` which to my understanding is a nonsensical pin mode.

Additionally, the body of the function forced the pin to be a floating input, but EXTI works with pins in any mode, so the original mode should persist.


usbalbin:
Thanks :)

PR #140, opened 2024-10-28
Fix warnings introduced with rust 1.82

usbalbin:
This fixes the warnings about irrefutable `if let` patterns introduced in rustc 1.82. However the fix as implemented in 9b27404 more or less raises the effective MSRV to 1.82...

We don't really have a MSRV as far as I know but perhaps 9b27404 should wait a while.


usbalbin:
I'll merge this now and we can consider actually using the shiny new features of rustc 1.82 at some point later..

PR #102, opened 2023-12-16
[(maybe?)DON'T MERGE] Update pac

usbalbin:
#96 requires updates to the pac. However the new pac changes register blocks member fields to access methods.

Almost all this PR does is adding paranthesis to things like
```diff
- dac.dac_cr.modify(|_, w| w.$en().clear_bit());
+ dac.dac_cr().modify(|_, w| w.$en().clear_bit());
```

This PR currently uses stm32g4 from the nightly repo, I assume at some point a new version of the pac is released. We can then merge this PR(after updating it to depend on that released pac).

I will use this as a base when working on #96 for now..

**Edit:**
Now using [stm32g4-staging](https://crates.io/crates/stm32g4-staging)

PR #142, opened 2024-11-08
Update stm32g4 (#102)

usbalbin:
Same as #102 but targeted at the new `staged-pac` branch @AdinAck


usbalbin:
I will merge this and we can continue from there?

AdinAck:
Sounds good!

PR #143, opened 2024-11-08
Fully Enforced Cordic

AdinAck:
With the new `from_*` register accessor methods, the fully type enforced Cordic implementation can be merged!


AdinAck:
Woops, wrong branch :/

Issue #145, opened 2024-11-09
DMA stream/channels


burrbull:
Why DMA channels named as `Stream`.
It is terminology from F4/H7.

Compare.

H7:
![изображение](https://github.com/user-attachments/assets/952f4d59-0738-4e8e-9dd4-339a4415e558)

G4:
![изображение](https://github.com/user-attachments/assets/de120803-4bb4-4af0-aab6-085b11d5e038)

usbalbin:
So in H7 it is stream 0-7, while G4 is channel 1-8?

burrbull:
Yes, also channels have execution priority when streams are more independent.

PR #147, opened 2024-11-11
DMA - rename stream to channel

usbalbin:


usbalbin:
@burrbull , does this fix #145 ?

PR #148, opened 2024-11-21
Allow selecting any clock source when freezing rcc configuration

liamkinne:
Currently PCLK is is statically configured when the constructor for the FDCAN peripheral is called. This pull request adds selecting the clock source.

This also means the default clock source is HSE (the reset value) rather than PCLK. The `can-echo` example doesn't need to be updated as it assumes PCLK = HSE.

### DIFF
--- a/src/can.rs
+++ b/src/can.rs
@@ -11,19 +11,6 @@ mod sealed {
     pub trait Rx<CAN> {}
 }
 
-/// Select an FDCAN Clock Source
-#[allow(clippy::upper_case_acronyms)]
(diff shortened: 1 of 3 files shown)


liamkinne:
@usbalbin once you are happy with this PR, I can do the same for all of the other peripheral clock sources.

usbalbin:
LGTM

liamkinne:
I will do the other clock configs in another PR. Happy for this to be merged.

usbalbin:
Thank you!

PR #157, opened 2024-12-21
Implement Observable POC

usbalbin:
Alternative to #138

@AdinAck is this somewhere close to what you discribed in #138 ?

### TODO
* Test
  * [x] ADC
  * [x] Comp
  * [x] DAC
  * [x] GPIO
  * [x] Opamp

### DIFF
--- a/examples/comp_w_dac.rs
+++ b/examples/comp_w_dac.rs
@@ -15,6 +15,7 @@ fn main() -> ! {
     use hal::delay::SYSTDelayExt;
     use hal::gpio::GpioExt;
     use hal::rcc::RccExt;
+    use hal::stasis::Freeze;
     use hal::stm32;
     use stm32g4xx_hal as hal;
 
@@ -29,15 +30,16 @@ fn main() -> ! {
     // Set up DAC to output to pa4 and to internal signal Dac1IntSig1
     // which just so happens is compatible with comp1
     let dac1ch1 = dp.DAC1.constrain((gpioa.pa4, Dac1IntSig1), &mut rcc);
-    let mut dac = dac1ch1.calibrate_buffer(&mut delay).enable(&mut rcc);
+    let dac = dac1ch1.calibrate_buffer(&mut delay).enable(&mut rcc);
+    let (mut dac, [dac_token]) = dac.freeze();
 
     let (comp1, _comp2, ..) = dp.COMP.split(&mut rcc);
     let pa1 = gpioa.pa1.into_analog();
 
     // Set up comparator with pa1 as positive, and the DAC as negative input
     let comp = comp1.comparator(
-        &pa1,
-        &dac,
+        pa1,
+        dac_token,
         comparator::Config::default().hysteresis(comparator::Hysteresis::None),
         &rcc.clocks,
     );
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -9,6 +9,7 @@
 */
 
 //use crate::dma::traits::PeriAddress;
+use crate::stasis;
 pub use crate::time::U32Ext as _;
 use crate::{
     dma::{mux::DmaMuxResources, traits::TargetAddress, PeripheralToMemory},
@@ -51,12 +52,16 @@ impl Vref {
         Self::sample_to_millivolts_ext(sample, VDDA_CALIB, config::Resolution::Twelve)
     }
 }
+impl stasis::Freeze for Vref {}
 
 /// Vbat internal signal, used for monitoring the battery (if used)
 pub struct Vbat;
+impl stasis::Freeze for Vbat {}
 
 /// Core temperature internal signal
 pub struct Temperature;
+impl stasis::Freeze for Temperature {}
+
 impl Temperature {
     /// Precompute the inverse of `VTEMP_CAL_VREFANALOG`, in volts,
     /// for floating point calculations
@@ -131,13 +136,38 @@ impl Temperature {
     }
 }
 
+// TODO: Is there any way to avoid this wrapper to overcome the orphan rule
+/// Wrapper to side step orphan rule
+pub struct Ad<T>(T);
+macro_rules! adc_channel_helper {
+    ($adc:ident, $chan:expr, $r:ty, $($a:ident),*) => {
+        impl<$($a,)*> Channel<Ad<stm32::$adc>> for stasis::Entitlement<$r> {
+            type ID = u8;
+            fn channel() -> u8 {
+                $chan
+            }
+        }
+
+        impl<const N: usize, $($a,)*> Channel<Ad<stm32::$adc>> for stasis::Frozen<$r, N> {
+            type ID = u8;
+            fn channel() -> u8 {
+                $chan
+            }
+        }
+
+        impl<$($a,)*> Channel<Ad<stm32::$adc>> for $r {
+            type ID = u8;
+            fn channel() -> u8 {
+                $chan
+            }
+        }
+    };
+}
+
 macro_rules! adc_pins {
     ($($pin:ty => ($adc:ident, $chan:expr)),+ $(,)*) => {
         $(
-            impl Channel<stm32::$adc> for $pin {
-                type ID = u8;
-                fn channel() -> u8 { $chan }
-            }
+            adc_channel_helper!($adc, $chan, $pin,);
         )+
     };
 }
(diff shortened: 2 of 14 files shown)


src/observable.rs
```diff
@@ -0,0 +1,75 @@
+use crate::Sealed;
+use core::marker::PhantomData;
+
+pub trait IntoObservationToken: Sized + crate::Sealed {
+    type Peripheral;
+    fn into_ot(self) -> ObservationToken<Self::Peripheral>;
+}
```
usbalbin, line 7: I had to do this since we can not do this the trait impl rules

```rust
impl<T: Observable> Into<ObservationToken<T>> for T {
    fn into(self) -> ObservationToken<T> {
        ObservationToken {
            _p: PhantomData
        }
    }
}
```

do you know if there is a better way?

AdinAck: The most recent version of this concept (with a few name changes and being used in a brand new context) is [here](https://github.com/AdinAck/proto-hal/blob/main/proto-hal/src/stasis.rs). The simple solution to this is to implement `From` rather than `Into`. This is a good idea in general (save some rare edge cases).

```rust
impl<P> From<P> for ObservationToken<P>
where
    P: Observable,
{
    fn from(_: P) -> Self {
        Self { _p: PhantomData }
    }
}
```

Another option - which eliminates the need for calling `.into()` to directly pass a peripheral as an observation token - is to create a new trait called `ObservationLock` which is implemented by all `P: Observable` and `ObservationToken`s. Interfaces would thus hold `impl ObservationLock`s rather than concrete `ObservationToken`s. This is described in [this](https://gist.github.com/AdinAck/c5713baf8b92d7075e10a9e03591569a) gist.

The former is a solution I am putting to use right now in [proto-hal](https://github.com/AdinAck/proto-hal/tree/main), the latter I suspect is a better solution, but have not implemented quite yet.

Hope this helps!

src/observable.rs
```diff
@@ -0,0 +1,75 @@
...
+
+/// A trait providing an interface to make peripherals observed
+///
+/// See [`Observed`] and [`ObservationToken`]
+pub trait Observable: Sized {
+    fn observe<const N: usize>(self) -> (Observed<Self, N>, [ObservationToken<Self>; N]) {
+        (
+            Observed { peripheral: self },
+            core::array::from_fn(|_| ObservationToken { _p: PhantomData }),
```
AdinAck, line 44: I did some testing and determined this does not introduce any overhead. I recall that being a concern.

AdinAck (commented):
This looks great! Have you run any integration tests?

src/observable.rs
AdinAck: Maybe the `ObservationLock` trait is unneeded since the interfaces can accept an `Into<ObservationToken<P>P>`. Clever!

src/observable.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl<P, const OBSERVER_COUNT: usize> Observed<P, OBSERVER_COUNT> {
+    /// Release the observation of this peripheral
+    ///
+    /// This returns the underlaying perpheral type. Since it is no longer
+    /// observed, you are once again free to do what you want with it.
+    pub fn release(self, _data: [ObservationToken<P>; OBSERVER_COUNT]) -> P {
+        self.peripheral
+    }
+}
```
AdinAck, line 27: You should definitely constrain `P` to `Observable`!

AdinAck (changes requested):
Every occurrence of `P` should be `P: Observable`.

src/observable.rs
usbalbin: If we want the user to be able to destruct the consumer of the `ObservationToken` how does this affect things?

If the user passed an actual owned peripheral than would it not be best to just **not** convert it and let the consumer store the peripheral as is until it is destructed?

In that case would `ObservationLock` make more sense since that better signals what we are doing?

AdinAck: Ah yes! *That* was why `ObservationLock` was good. Since the peripherals would store an `impl ObservationLock`, the `release()` would return back whatever was originally passed in.

src/observable.rs
usbalbin: Sorry for dumb question, but why? The only way to construct an `Observed` is with a `P` that is `Observable`, right? Just curious :)

src/observable.rs
usbalbin: Nice, thanks!

src/observable.rs
AdinAck: Not dumb at all. In all of these cases we *want* `P` to be `Observable`. It may be true that right now it just so happens that the only way to encounter these interfaces is when `P` is `Observable`, but I would consider that to be a weak invariance given that we are actively making significant changes that could silently break things. Having these additional constraints may catch other design mistakes.

Up to you, but I prefer erring on the side of caution at the expense of additional verbosity.

usbalbin: Thanks! Makes sense

examples/comp_w_dac.rs
```diff
@@ -59,7 +61,7 @@ fn main() -> ! {
     // * 0V   at p1 =>   0% duty
     // * VDDA at p1 => 100% duty
     loop {
-        dac.set_value(val);
+        dac.as_mut().set_value(val);
```
usbalbin, line 64: Should we even go as far as implementing `Deref{Mut}` for `Observed`?

AdinAck: Yep, `Deref` will be a really good QoL feature.

usbalbin:
> This looks great! Have you run any integration tests?

Thanks :) No, no tests at all yet

src/observable.rs
usbalbin: > The former is a solution I am putting to use right now in [proto-hal](https://github.com/AdinAck/proto-hal/tree/main)

Very interesting!

src/adc.rs
```diff
@@ -136,7 +137,9 @@ impl Temperature {
 macro_rules! adc_pins {
     ($($pin:ty => ($adc:ident, $chan:expr)),+ $(,)*) => {
         $(
-            impl Channel<stm32::$adc> for $pin {
+            impl<P> Channel<stm32::$adc> for P
+                where P: stasis::EntitlementLock<Resource = $pin>
+            {
```
usbalbin, line 142: This does not work. Both `Channel` and `EntitlementLock` are from outside this crate. There are also errors about conflicting implementations with the other opamps below.

AdinAck: I think the simplest solution is to make a local trait like `AdcChannel` which requires `embedded_hal::adc::Channel` and then it would work.

I'm not very familiar with what you're doing here though. I'll think on it.

You could also do concrete implementations since this is macro generated anyway.

usbalbin: > You could also do concrete implementations since this is macro generated anyway.

Yeah that would probably be the simplest solution. I'll give it a go

usbalbin:
@burrbull do you have any thoughts on this as a concept?

TLDR(please @AdinAck correct me where I am wrong here):

There are several cases where it is useful to have multiple observers of the same resource while preventing one observer from reconfiguring the resource in a way not compatible with the others.

For example a gpio pin may be "observed" by a comparator while still being usable for AD readings at the same time. However we want to prevent the user from reconfiguring the pin to something other than Analog mode.

Another example would be a Dac. We want it to be observable from the comparator and/or opamp while we still have access to change its output.

This proposes to solve this by introducing a trait `Freeze` that enables the user to freeze a resource that implements it. By freezing the resource you get a `Frozen<YourResource, N>` and an array of `N` number of `Entitlement<YourResource>`  which are proof that the resource exists and is indeed frozen in the expected state. Observers like the comparator or opamp may then take `Entitlement<YourResource>` instead of YourResource as inputs.

By recombining all `N` `Entitlement` with the `Frozen`, the resource may once again be unfrozen(defrosted?)

AdinAck:
You are exactly right!

Additional background: this is exactly what references are, the compiler just automatically tracks the "recombination" (dropping of references) and forbids moving or dropping the resource if a reference to it exists. We are electing not to use references for two reasons:
1. Dealing with lifetimes.
2. These resources are of zero size, and we leverage that to produce really good compiler optimizations. Currently, references to ZSTs are not ZSTs themselves so we cannot express composite structures of these resources as ZSTs as well.

The drawbacks of this approach is we need to explicitly dispatch "entitlements" which are our reference analog, and explicitly recombine them, which is a bit of a hassle.

usbalbin:
@AdinAck do you have any suggestions for how to move forward with this? Are you planning to release your stasis thing so we can depend on it from stm32g4xx-hal (and later I would probably want to use it for [usbalbin/stm32-hrtim](https://github.com/usbalbin/stm32-hrtim) too). Or do you think I should have a local copy in stm32g4xx-hal for now? :)

AdinAck:
I am still actively developing proto-hal and haven't reached a stable release point yet, so I wouldn't want my changes to affect you. I think a local copy makes sense, for now :)

usbalbin:
Makes sense, thanks

Is it ok if I copy your stasis.rs? [8df2caf](https://github.com/stm32-rs/stm32g4xx-hal/pull/157/commits/8df2caf1f549cd6cb58854470a887cb6fabae10d)

AdinAck:
Yep!

usbalbin:
> Yep!

Awesome! Thank you!

tests/nucleo-g474_w_jumpers.rs
```diff
@@ -0,0 +1,316 @@
+#![no_std]
```
usbalbin, line 1: @AdinAck 
> This looks great! Have you run any integration tests?

I have now added some more tests. These ones require a jumper(I took the one at CN12 at the top left corner) when using a Nucleo-G474 between A1 and A2. But I do believe that single modification opens up for some pretty interesting self tests

Issue #112, opened 2024-01-04
uart-dma example with changing data length


usbalbin:
The uart-dma example does not show how to send data with different length using dma.

Is this not supported by this crate or is it even a hardware limitation?

usbalbin:
To me I think it would have made more sense if the uart and dma stream were combined into a "uart-with-dma" rather than a "dma-stream-thing that happens to wrap an uart".

So instead of
```rust
let mut transfer =
    streams
        .0
        .into_memory_to_peripheral_transfer(tx.enable_dma(), &mut tx_buffer[..], config);
```
I would want
```rust
let mut tx: Tx<UART2, PIN, DMA> = tx.enable_dma(streams.0, &mut tx_buffer[..]);
```
since `Tx<UART2, PIN, DMA>` can then impl `core::fmt::Write`, set senseful dma configs automatically etc since it has more knowledge of how the dma channel will be used.

Any thoughts? :)

PR #162, opened 2025-01-07
Simplify DMA impl removing most macros

usbalbin:
Simplify DMA impl by not using macros now that the pac has been improved using the same pac types across the two DMAs

### DIFF
--- a/src/dma/channel.rs
+++ b/src/dma/channel.rs
@@ -346,7 +346,7 @@ impl<I: Instance, const N: u8> Channel for C<I, N> {
     fn get_half_transfer_flag(&self) -> bool {
         //NOTE(unsafe) Atomic read with no side effects
         let dma = unsafe { &*I::ptr() };
-        dma.isr().read().htif(N as u8).bit_is_set()
+        dma.isr().read().htif(N).bit_is_set()
     }
 
     #[inline(always)]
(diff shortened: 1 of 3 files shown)


usbalbin:
@AdinAck do you have any thoughts on this? I have quite limited experiance with DMA, but I get the feeling the DMA story of this crate is less than ideal as is.

AdinAck:
Just by coincidence a few days ago I used it for the first time and I was very frustrated very quickly, so suffice it to say I am happy for any improvements, and am interested in helping with that effort.

I need to think a lot harder before I propose any significant design changes, this is something I have actively skipped thinking about in my HAL design journey. But I think soon it will be time.

In general I do believe *traitifying* everything as much as possible is the correct direction. For example, when designing abstract structures intended for reuse, if I want let's say to stream some bytes out over UART via DMA, I of course will hold something of type `Transfer<{lot's of generics}>`.

This is bad because then my abstraction needs to hold many generics related to lower implementation details.

I don't care what generics the `Transfer` has, I just want any kind of `Transfer`.

Embassy's solution is to simply erase this type information. They went with that approach because tasks cannot accept generics, so the full type signature would need to be elaborated, defeating the purpose of lifting methods into a trait. I don't like this approach because type-erasure forbids compile-time evaluation and opens the door for incurred runtime costs.

My idea for a good solution would be if `Transfer<...>` implemented a trait let's call `TransferAgent` (horrible name, doesn't matter) which houses all of the transfer methods, i.e. `start`, `peek_buffer`, `pause`, etc. That way abstractions simply hold an `impl TransferAgent`. Additionally, since the types and their respective generics are all still resolved at compile-time, no runtime cost is incurred for differing implementations.

If lower-level specification is *really* necessary, the generics could be concretely specified via associated types of the trait: `impl TransferAgent<Stream = hal::dma::streams::Stream5>`.

I believe many HAL components would benefit from this design. I'm just spitballing here though, I have not tested this approach quite yet, although I intend to do so soon.

I don't know if this is too much for what you were asking.

usbalbin:
Thanks for the answer :)

> I am happy for any improvements, and am interested in helping with that effort.

Great:)

> I don't care what generics the Transfer has, I just want any kind of Transfer.

Makes sense 

There is also #112

usbalbin:
I'll merge this now as is and we can continue with further improvements in future PRs

PR #163, opened 2025-01-08
defmt Format for RCC Clocks

AdinAck:
Some initial `Format` derives for RCC clocks, really lots of structures should derive `Format`.

I also bumped `fugit`.

The idiomatic `defmt` feature across lots of crates seems to be the name "defmt" itself, so I also changed the `defmt-logging` feature to just `defmt`.


usbalbin:
Great! Thanks :)

PR #164, opened 2025-01-08
Export PwrExt In Prelude

AdinAck:
I noticed the `PwrExt` re-export was missing from `prelude`.

On a side note, why are so many exports commented out? Is this a regression that slipped by?


usbalbin:
Seems to have been that way for quite some time https://github.com/stm32-rs/stm32g4xx-hal/blame/main/src/prelude.rs

PR #167, opened 2025-01-12
Staged pac

usbalbin:


usbalbin:
@AdinAck , @burrbull , @liamkinne do you see any reason to keep the separate branch with the changes for the new pac, or should we just merge this?

AdinAck:
The only constraint on my end is that `cordic` relies on `from_modify`, which was previously only in the staged PAC. If that has been merged into the normal PAC then it should be fine!

usbalbin:
As far as I know the [normal PAC](https://crates.io/crates/stm32g4) still has not released any new versions since about 2 years ago. I do not know the story there, but @burrbull has been very quick with releasing new versions of [staging PAC](https://crates.io/crates/stm32g4-staging) when new features have been merged. (thanks a lot)

Since the last "normal PAC"-release there has been a whole lot of bug fixes, lots of deduplications of types, arrayification of fields, making more things safe, and your `from_modify` (and probably a whole lot of more things that I do not understand) etc that all in all allow a lot of nice cleanup in the hal.

I made the `staged-pac`-branch in this hal to target the "staging PAC" releases of the pac so we can start using all the new improvements. At this point the regular master branch is starting to get outdated with issues popping up like #184 and I believe more that I have forgotten which are solved in `staged-pac`.

So, in short, any reason to not just merge this `staged-pac`-branch into `master` and thus have `master` depend on `staging PAC`-releases of the pac?

burrbull:
The story is simple. @adamgreig is busy all time. Let's ask him on next meeting. Maybe something has been changed.

usbalbin:
> The story is simple. @adamgreig is busy all time. [...]

I understand, and that is fine :)  (Also let me know if there is anything I can help with)

> Let's ask him on next meeting. Maybe something has been changed.

No rush on my account :). I see [stm32f4xx-hal is using the staging-pac](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/Cargo.toml#L68) do you think we should merge this? (I am also more than happy to wait)

burrbull:
> No rush on my account :). I see [stm32f4xx-hal is using the staging-pac](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/Cargo.toml#L68) do you think we should merge this? (I am also more than happy to wait)

Yes, I use `staging` branch in f4xx-hal during a year. So you know my position.

burrbull:
Good time to merge in the main :)

usbalbin:
I agree :)

PR #168, opened 2025-01-13
[TRACKING] Derive `defmt::Format`

AdinAck:
This PR tracks the addition of `defmt::Format` derives on various HAL structures.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -18,34 +18,43 @@ pub trait GpioExt {
 }
 
 /// Input mode (type state)
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct Input<MODE> {
     _mode: PhantomData<MODE>,
 }
 
 /// Floating input (type state)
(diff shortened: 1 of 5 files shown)


src/gpio.rs
```diff
@@ -267,6 +279,7 @@ macro_rules! gpio {
             }
 
             /// Partially erased pin
+            #[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
burrbull, line 282: https://github.com/stm32-rs/stm32f4xx-hal/blob/585dd0f30443c45a1be9f6483fd480b141d819a2/src/gpio/partially_erased.rs#L40

AdinAck: Ah interesting, the derived formatting string is not good?

AdinAck: Oh this is in a detached commit. What is this?

burrbull:
`stm32g4-staging` now depends on `defmt` of version 1

PR #179, opened 2025-01-28
Add Tests

usbalbin:
Thought I would try to experiment a bit with adding tests that could run an a Nucleo or similar without any ~~/with very little~~ external connections.

From what I can tell the input registers are supposed to be usable even after the pin has been put into alternate mode. So in my mind I should for example be able to verify that the PWM is working by measuring the time on and time off.

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -95,7 +95,7 @@ mod tests {
     }
 
     #[test]
-    fn foo() {
+    fn pwm() {
         use super::*;
 
         // TODO: Is it ok to steal these?
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -4,27 +4,51 @@
 #[path = "../examples/utils/mod.rs"]
 mod utils;
 
-use utils::logger::println;
+use utils::logger::debug;
 
 use core::ops::FnMut;
 use core::result::Result;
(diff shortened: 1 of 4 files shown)


tests/tests.rs
```diff
@@ -0,0 +1,336 @@
...
+
+        /*for i in (0..=4095).step_by(10) {
+            dac.set_value(i);
+            delay.delay_ms(1);
+            defmt::println!("i: {}, is_pax_low: {}", i, gpioa.idr().read().bits());
+        }
+
+        delay.delay_ms(100);
+        assert!(!is_pax_low(&gpioa, 4)); // TODO: <---- Why does this not work?
+        */
```
usbalbin, line 296: Any reason why this does not work?

The pin seems to thinkt that it is low no matter what value I set.

Will try to measure and see what voltage the pin in fact is

usbalbin:
Tested by running `cargo test --features stm32g474,defmt,cordic --tests -- --chip stm32g474RETx` on a [NUCLEO-G474RE](https://www.st.com/en/evaluation-tools/nucleo-g474re.html#overview) and currently passes all tests

usbalbin:
Tested again after rebase

PR #182, opened 2025-01-30
Enable *and* reset gpio in split method

usbalbin:
@AdinAck and @burrbull , do you believe this is the actual fix for #180 ?

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -259,7 +259,7 @@ macro_rules! gpio {
                 fn split(self, rcc: &Rcc) -> Parts {
                     $GPIOX::enable(&rcc.rb);
                     $GPIOX::reset(&rcc.rb);
-                    
+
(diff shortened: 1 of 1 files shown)


src/gpio.rs
```diff
@@ -14,7 +14,7 @@ pub trait GpioExt {
     type Parts;
 
     /// Splits the GPIO block into independent pins and registers
-    fn split(self, rcc: &mut Rcc) -> Self::Parts;
+    fn split(self, rcc: &Rcc) -> Self::Parts;
```
usbalbin, line 17: Also, unrelated to #180, since `Enable::enable` and `Reset::reset` use bit banding, we do not need exclusive access to the rcc right?

usbalbin: Will remove this and perhaps do that later

PR #183, opened 2025-02-04
Update for new pac

usbalbin:
This is the changes needed to go from `stm32g4-staging = "0.19.0"` to stm32-rs/stm32-rs#1156

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -1839,26 +1839,11 @@ macro_rules! adc {
                 pub fn set_channel_input_type(&mut self, df: config::DifferentialSelection) {
                     self.config.difsel = df;
 
-                    self.adc_reg.difsel().modify(|_, w| {w
-                        .difsel0().bit(df.get_channel(0).into() )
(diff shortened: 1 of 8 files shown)


src/adc.rs
```diff
@@ -1840,25 +1840,25 @@ macro_rules! adc {
                     self.config.difsel = df;
 
                     self.adc_reg.difsel().modify(|_, w| {w
```
burrbull, line 1842:
```rust
                   self.adc_reg.difsel().modify(|_, w| {
                       for i in 0..18 {
                           w.difsel(i).bit(df.get_channel(i).into())
                       }
                   });
```
Not sure how it is optimized.

usbalbin: Thanks!

burrbull:
> If you would like to release that,

A month ago.

usbalbin:
> > If you would like to release that,
> 
> A month ago.

Thanks! :)

PR #187, opened 2025-04-17
fix OPAMP opaintoen in PGA mode using internal output

Jonarw:
`OPAMPX_CSR.OPAINTOEN` is incorrectly configured as 0 (OutputPin) even when the Opamp is configured using `InternalOutput` like this:

```
    opamp1.pga_external_bias(
        pins.pa1.into_analog(),
        pins.pa3.into_analog(),
        InternalOutput,
        Gain::Gain64,
    );
```

This PR fixes this by calling `configure_pga` with `output_enable=false` when `InternalOutput` is used.

### DIFF
--- a/src/opamp.rs
+++ b/src/opamp.rs
@@ -813,7 +813,7 @@ macro_rules! opamps {
                     gain: Gain,
                 ) -> Pga<$opamp, $non_inverting, InternalOutput>
                 {
-                    $opamp::configure_pga(InternalOutput, gain, PgaMode::Pga, true)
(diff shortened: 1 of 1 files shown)


usbalbin:
Nice catch, thanks!

Have you tested this?

Jonarw:
Yes, I've tested this on an STM32G431.

usbalbin:
Thank you for this!

usbalbin:
The failed tests are unrelated. I will try to address them at a later time

PR #199, opened 2025-05-12
Demacrofy ADC

### LINKED ISSUE #124: ADC calibration should be done after setting the clock mode
Hi,

I noticed on my board nucleo-G474-RE when connecting an ADC to ground with RC filter, that the readings were off by 50mV. I noticed that calibrating the ADC before setting the clock config (which the enable function of the HAL also does) results in wrong calibration value (almost max register value). Which resulted in 50 bits offset being applied to my readings. The ADC doesn't work correctly when the clock is higher than the maximum (60MHz), so before the clock mode settings are set the clock might be AHB clock. Calibration should be done before ADEN=1 but after setting the clock mode in the common ADC registers.

Easy fix might be to swap around calibrate_all() and apply_config() calls to calibrate after setting clocks (And other config options, but before enabling the ADC).

### LINKED ISSUE #181: ADC - Unsafe reset parameter
In light of #180 is `reset: bool` in https://github.com/stm32-rs/stm32g4xx-hal/blob/main/src/adc.rs#L2135 really safe to expose considering the user may pass `false`?. Also is this really a good way to do things since ADC1 and ADC2 share the same RCC enable and reset bit [PAC](https://stm32-rs.github.io/stm32-rs/STM32G474.html#RCC:AHB2RSTR:ADC12RST) [HAL](https://github.com/stm32-rs/stm32g4xx-hal/blob/116e6d6d6777cf9ed32981fa5fb9eaf058f63f6f/src/rcc/enable.rs#L79). Same thing with [ADC3-ADC5](https://stm32-rs.github.io/stm32-rs/STM32G474.html#RCC:AHB2RSTR:ADC345RST).

In other words if the user does
```rust
let adc1 = dp.ADC1.claim(..., reset=false); // Might not have waited long enough after rcc enable to register access
```
or
```rust
let adc1 = dp.ADC1.claim(..., reset=true);
let adc2 = dp.ADC2.claim(..., reset=true); // <--- ADC1 will be reset here
```

both seems, in the very least, like huge footguns to me.

Slightly similar situation with FDCAN1-3 sharing the [same enable and reset bits](https://stm32-rs.github.io/stm32-rs/STM32G474.html#RCC:APB1RSTR1:FDCANRST). However the [hal](https://github.com/stm32-rs/stm32g4xx-hal/blob/main/src/can.rs#L43) does not seem to reset them at all on init. Or wait for that matter

usbalbin:
Demacrofy ADC and move clock logic that is common for the group from `Adc<...>` to `AdcCommon<...>`

Fixes #181
Fixes #124 (most likely)


usbalbin:
cc @burrbull

src/adc/mod.rs
```diff
@@ -0,0 +1,1580 @@
...
+            }
+        });
+
+        let channel = CHANNEL::channel();
+
+        //Set the channel in the right sequence field
+        match sequence {
+            config::Sequence::One => self
+                .adc_reg
+                .sqr1()
```
burrbull, line 833: `.sqr(n)` should be possible to use here. Same for `smpr`

src/pwm.rs
```diff
@@ -184,7 +184,6 @@ use crate::stm32::RCC;
 ))]
 use crate::stm32::TIM20;
 #[cfg(any(
-    feature = "stm32g471",
```
burrbull, line 187: What is wrong with `stm32g471`?

usbalbin: I can not for the life of me even find a datasheet for it. And I can not seem to find where to buy it either. I have checked mouser, digikey and lcsc. From what I can tell it is only mentioned in the reference manual. It does not even exist on [ST's site](https://www.st.com/en/microcontrollers-microprocessors/stm32g4x1.html)

AdinAck: The 71 seems to be the only variant without an AES unit. It probably fell out of favor and as such was discontinued. The datasheet isn't even distributed anymore.

src/adc/mod.rs
usbalbin: Thanks!

PR #201, opened 2025-05-26
Set dac hfsel

usbalbin:


usbalbin:
@burrbull is e4e7ce0 how it should be done?

burrbull:
Yes. But you should also enable it in repo "Settings".

usbalbin:
Thank you, lets try it :)

PR #209, opened 2025-05-28
Cleanup tests

usbalbin:


tests/nucleo-g474.rs
```diff
@@ -91,27 +86,28 @@ mod tests {
...
             let gpioa = unsafe { &*GPIOA::PTR };
-            gpioa.pupdr().modify(|_, w| w.pupdr8().pull_up());
+            gpioa.pupdr().modify(|_, w| w.pupdr(pin_num).pull_up());
         }
 
         pin.set_high();
         delay.delay(1.millis()); // Give the pin plenty of time to go high
         assert!(pin.is_high());
         {
             let gpioa = unsafe { &*GPIOA::PTR };
```
usbalbin, line 101: @burrbull Are we allowed to keep around this reference so its lifetime overlaps with the usage of `pin`? I added the scope here to avoid that but if allowed then we could create the reference once and keep that instead of all the blocks

PR #210, opened 2025-05-30
rcc::Enable and cleans

burrbull:


src/pwm.rs
```diff
@@ -1148,9 +1147,8 @@ macro_rules! tim_hal {
                 PINS: Pins<$TIMX, T, U>,
             {
                 unsafe {
-                    let rcc_ptr = &(*RCC::ptr());
-                    $TIMX::enable(rcc_ptr);
-                    $TIMX::reset(rcc_ptr);
+                    $TIMX::enable_unchecked();
```
usbalbin, line 1150: Now that Rcc derefs to RCC, could we just pass rcc here instead of rcc_ptr?

usbalbin: In that case we would not need to use the `_unchecked` variants, right?

burrbull: `enable_unchecked` is just synonym for:
```rust
PER::enable(&mut RCC::steal());
```

burrbull: > we do have access to it as the parameter `rcc`?

Sorry. This is what I've missed.

burrbull: > Sorry for being unclear :)

This is still unclear for me. What exactly do you expect from this PR? Adding `rcc: &mut Rcc` where possible?
Like in https://github.com/stm32-rs/stm32f4xx-hal/pull/836 ?

usbalbin: For this PR i would prefer avoiding `_unchecked();` when we already have `rcc: &mut Rcc` if possible.

But adding `rcc: &mut Rcc` in more places as in https://github.com/stm32-rs/stm32f4xx-hal/pull/836 (and figuring out a solution for the usb case) can be left for a future PR.

burrbull: :laughing: :laughing: :laughing: I didn't even think to look if `rcc` is already present in these functions.

usbalbin: Again sorry for the confusion :) Thanks a lot this looks very nice now!

usbalbin (approved):
All tests pass on my nucleo-g474

burrbull:
ok

PR #212, opened 2025-06-22
Fixes incorrect DMA ISR flag used for transfer error status

boondocklabs:
Simple fix to use correct ISR status flags:

* `get_transfer_error_flag()` change from `TCIF` flag to `TEIF`.
* `clear_transfer_error_interrupt()` change from `CTCIF` flag to `CTEIF`.


usbalbin:
Thank you!

PR #213, opened 2025-06-23
DAC Improvements - Dual DAC support and DMA

boondocklabs:
* Add `DualDac` with simultaneous channel writes, with DMA support
* Added `SampleFormat` trait and impls supporting Q1.7, Q1.11, Q1.15 signed and unsigned 8/11b types, left/right alignment
* Added `DacTriggerSource` and `DacIncrementSource` traits for selecting external trigger sources with DAC instance constraints
* Added DMA support to existing single channel DACs

Here's an example of using a circular phase angle buffer which is DMA'd into CORDIC in SinCos mode, with DualDac configured to DMA samples directly from CORDIC RDATA register in Q1.15 format, and triggered by Tim6 at 1MHz. This produces a quadrature output on both DAC1 channels, with zero CPU.

DualDac Quadrature CORDIC DMA

### DIFF
--- a/src/dac/dual.rs
+++ b/src/dac/dual.rs
@@ -1,0 +1,206 @@
+#![deny(missing_docs)]
+
+//! Dual DAC
+//!
+//! This module provides access to the dual DAC channels of the STM32G4 series microcontrollers.
+//! It allows for simultaneous access to both DAC channels and supports DMA transfers.
+//!
+
+use core::marker::PhantomData;
+
+use embedded_hal::delay::DelayNs;
+
+use crate::dac::{
+    format::SampleFormat, trigger::DacTriggerSource, DacCh, DacChannel, Disabled, Enabled,
+    Instance, M_EXT_PIN,
+};
+use crate::dac::{hfsel, DacExt as _, Pins};
+
+use crate::dac::format::{self, ToDac as _};
+use crate::dma::mux::DmaMuxResources;
+use crate::dma::traits::TargetAddress;
+use crate::dma::MemoryToPeripheral;
+use crate::rcc::Rcc;
+
+/// Extension trait for [`Instance`] to create a [`DualDac`] from the given pins and RCC.
+pub trait DualDacExt: Instance + Sized {
+    /// Create a dual DAC instance from the given pins and RCC.
+    ///
+    /// DAC calibration is performed before enabling the DAC, so
+    /// a DelayNs implementation is required.
+    ///
+    /// This uses the dual hold registers (DHR12xD, DHR8RD) and
+    /// can write to both DAC channels simultaneously with DMA support.
+    fn into_dual<F: SampleFormat, PINS>(
+        self,
+        pins: PINS,
+        rcc: &mut Rcc,
+        delay: &mut impl DelayNs,
+    ) -> DualDac<Self, F, Disabled>
+    where
+        PINS: Pins<
+            Self,
+            Output = (
+                DacCh<Self, 0, M_EXT_PIN, Disabled>,
+                DacCh<Self, 1, M_EXT_PIN, Disabled>,
+            ),
+        >;
+}
+
+/// Dual DAC handle
+pub struct DualDac<DAC: Instance, F: SampleFormat, State> {
+    _ch1: DacCh<DAC, 0, M_EXT_PIN, State>,
+    _ch2: DacCh<DAC, 1, M_EXT_PIN, State>,
+    _phantom: PhantomData<F>,
+}
+
+impl<DAC: Instance, F: SampleFormat> DualDac<DAC, F, Disabled> {
+    /// Enable DMA double mode for the specified channel.
+    /// This creates a single DMA request for every
+    /// two external hardware triggers (excluding software triggers).
+    #[inline(always)]
+    pub fn enable_dma_double(&mut self, channel: u8, enable: bool) {
+        let dac = unsafe { &(*DAC::ptr()) };
+        if enable {
+            dac.mcr().modify(|_, w| w.dmadouble(channel).set_bit());
+        } else {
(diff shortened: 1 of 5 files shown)


src/dac/dual.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// This creates a single DMA request for every
+    /// two external hardware triggers (excluding software triggers).
+    #[inline(always)]
+    pub fn enable_dma_double(&mut self, channel: u8, enable: bool) {
+        let dac = unsafe { &(*DAC::ptr()) };
+        if enable {
+            dac.mcr().modify(|_, w| w.dmadouble(channel).set_bit());
+        } else {
+            dac.mcr().modify(|_, w| w.dmadouble(channel).clear_bit());
+        }
```
usbalbin, line 68: Correcct me if I am wrong but i think there might be a method `bit` that could be used here.
```suggestion
        dac.mcr().modify(|_, w| w.dmadouble(channel).bit(enable));
```

src/dac/format.rs
```diff
@@ -0,0 +1,119 @@
...
+    const CH2_SHIFT: u8 = 8;
+    const CH1_MASK: u32 = 0xFF;
+    type Scalar = I1F7;
+}
+
+/// Fixed point signed Q1.11 samples
+pub struct SampleQ11;
+impl SampleFormat for SampleQ11 {
+    const DEPTH: SampleDepth = SampleDepth::Bits12;
+    const ALIGNMENT: Alignment = Alignment::Left;
```
usbalbin: I have very limited experiance with fixed point formats, but should this really be identical to `SampleQ15`?

boondocklabs: Well, G4 does claim support for Q1.11 and Q1.15 (1 sign bit, 11 or 15 fractional bits) so I included them both as types, but Q1.15 is a truncation as the DAC only outputs 12 bits and discards the lower 4 bits, and the fixed crate doesn't have sub 16b types (both scalar types are I1F15, hence the same shift and mask). So yes, they are identical.

src/dac/dual.rs
```diff
@@ -0,0 +1,191 @@
...
+    _ch2: DacCh<DAC, 1, M_EXT_PIN, State>,
+    _phantom: PhantomData<F>,
+}
+
+impl<DAC: Instance, F: SampleFormat> DualDac<DAC, F, Disabled> {
+    /// Enable DMA double mode for the specified channel.
+    /// This creates a single DMA request for every
+    /// two external hardware triggers (excluding software triggers).
+    #[inline(always)]
+    pub fn enable_dma_double(&mut self, channel: u8, enable: bool) {
```
usbalbin, line 62: `channel: u8` vs `channel: DacChannel` below

src/dac/dual.rs
```diff
@@ -0,0 +1,191 @@
...
+            Output = (
+                DacCh<Self, 0, M_EXT_PIN, Disabled>,
+                DacCh<Self, 1, M_EXT_PIN, Disabled>,
+            ),
+        >,
+    {
+        let (ch1, ch2) = self.constrain(pins, rcc);
+
+        let ch1 = ch1.calibrate_buffer(delay);
+        let ch2 = ch2.calibrate_buffer(delay);
```
usbalbin, line 159: Is this ok to do before setting `hfsel`?

boondocklabs: RM0440 doesn't mention it specifically, and HFSEL is for data output register rate (copies from hold register to output), so I don't imagine it would affect calibration which just sets the DAC to midpoint output 0x800 and uses the buffer as a comparator internally against Vref/2 (RM0440 22.4.13).

I haven't had any issues with 24MHz HSE -> 168MHz PLL sysclock, and 1M DAC sample rate

usbalbin: Ok nice, that should probably be fine then.

boondocklabs:
> This is very nice!
> 
> Do you have any example of using the dual mode or dma that you can share? :)

I'll try and get an example pushed today. My current codebase depends on some other modifications for CORDIC DMA (which I'll also open PRs for), but I can make something standalone.

usbalbin:
Sounds great!

I'll go to sleep now, time zones.. wont be back for a bounch of hours. Thanks so far :)

src/dac/format.rs
AdinAck: We use `Q15` and `Q31` in the CORDIC peripheral as well. Maybe these should be coalesced to avoid repeat definitions?

AdinAck:
> ...My current codebase depends on some other modifications for CORDIC DMA (which I'll also open PRs for)...

Wow! 🤩 I am excited to see that!

boondocklabs:
> > This is very nice!
> > Do you have any example of using the dual mode or dma that you can share? :)
> 
> I'll try and get an example pushed today. My current codebase depends on some other modifications for CORDIC DMA (which I'll also open PRs for), but I can make something standalone.

I pushed a simple example. I can get a DMA example up at some point - maybe after the CORDIC DMA patches are in and can show an example of CORDIC direct to DAC via DMA.

src/dac/dual.rs
```diff
@@ -0,0 +1,192 @@
...
+    #[inline(always)]
+    pub fn enable_dma_double(&mut self, channel: DacChannel, enable: bool) {
+        let dac = unsafe { &(*DAC::ptr()) };
+        dac.mcr()
+            .modify(|_, w| w.dmadouble(channel as u8).bit(enable));
+    }
+
+    /// Enable DMA for the specified channel
+    #[inline(always)]
+    pub fn enable_dma(&mut self, channel: DacChannel, enable: bool) {
```
usbalbin, line 70: Why does `DacCh::enable_dma` consume self and turn it into a new type while this one does not?

Also, is it ok to have `set_channels` still available after dma has been enabled?

PR #217, opened 2025-07-06
Dyn comp invert

usbalbin:
From what I can tell there is nothing preventing us from changing the polarity of the comparator on the fly


burrbull:
It is better to use `Polarity` enum here instead of `bool`.

usbalbin:
> It is better to use `Polarity` enum here instead of `bool`.

Something like this e368fd8?

PR #220, opened 2025-07-21
Fix ADC eoc/eos interrupts

boondocklabs:
When enabling ADC interrupts using `Eoc::Conversion`, both end of sequence and end of conversion interrupts were being enabled, with no way to clear the end of sequence interrupt. When selecting `Eoc::Sequence`, just the end of conversion interrupt was being enabled.

This fixes the enabled interrupt flags in EOSIE, and adds a `clear_end_sequence_flag` method to clear the end of sequence flag.


usbalbin (approved):

usbalbin:
Nice, Thank you! :)

PR #221, opened 2025-07-22
Fix SysCfg hardfault, Add I2C Fast Mode Plus enable method

boondocklabs:
The `SysCfg` peripheral used bit banding to set the APB2 peripheral clock enable bit which fails on an assertion on a G431KBT (have not tested on other chips, but I suspect they would do the same). This takes a mutable reference to `Rcc` in `SysCfg::constrain` and uses safe accessors to enable the clock.

Updated and tested the button example which is all that uses SysCfg.

```[INFO ] Configuring PLL (stm32_foc stm32-foc/src/main.rs:132)
[INFO ] System clock frequency: 168000000 (stm32_foc stm32-foc/src/main.rs:138)
[DEBUG] Write 20007FB0 (stm32g4xx_hal stm32g4xx-hal/src/bb.rs:42)
[ERROR] panicked at /Users/fuzz/wave/stm32g4xx-hal/src/bb.rs:44:5:
assertion failed: (PERI_ADDRESS_START..=PERI_ADDRESS_END).contains(&addr) (panic_probe panic-probe-1.0.0/src/lib.rs:104)
Firmware exited unexpectedly: Multiple
Core 0
    Frame 0: HardFault_ @ 0x08006394
       /Users/fuzz/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cortex-m-rt-0.7.5/src/lib.rs:1103:1
    Frame 1: HardFault <Cause: Escalated UsageFault <Cause: Undefined instruction>> @ 0x08005ce2```

### DIFF
--- a/examples/button.rs
+++ b/examples/button.rs
@@ -52,17 +52,9 @@ fn main() -> ! {
 
     let mut dp = stm32::Peripherals::take().expect("cannot take peripherals");
 
-    // Workaround for RTT when using wfi instruction
-    // Enable the debug sleep bits in DBGMCU,
-    // then enable DMA peripheral clock in AHB1ENR
-    dp.DBGMCU.cr().modify(|_, w| {
-        w.dbg_sleep().set_bit();
-        w.dbg_stop().set_bit();
-        w.dbg_standby().set_bit()
-    });
-
     let mut rcc = dp.RCC.constrain();
 
+    // Workaround for RTT when using wfi instruction
     // Enable an AHB peripheral clock for debug probe with wfi
     rcc.ahb1enr().modify(|_, w| w.dma1en().set_bit());
 
--- a/examples/button.rs
+++ b/examples/button.rs
@@ -5,8 +5,7 @@ use stm32g4xx_hal::{
     //delay::{DelayExt, SYSTDelayExt},
     gpio::{self, ExtiPin, GpioExt, Input, SignalEdge},
     rcc::RccExt,
-    stm32,
-    stm32::{interrupt, Interrupt},
+    stm32::{self, interrupt, Interrupt},
     syscfg::SysCfgExt,
 };
 
@@ -52,7 +51,21 @@ fn main() -> ! {
     utils::logger::init();
 
     let mut dp = stm32::Peripherals::take().expect("cannot take peripherals");
+
+    // Workaround for RTT when using wfi instruction
+    // Enable the debug sleep bits in DBGMCU,
+    // then enable DMA peripheral clock in AHB1ENR
+    dp.DBGMCU.cr().modify(|_, w| {
+        w.dbg_sleep().set_bit();
+        w.dbg_stop().set_bit();
+        w.dbg_standby().set_bit()
+    });
+
     let mut rcc = dp.RCC.constrain();
+
+    // Enable an AHB peripheral clock for debug probe with wfi
+    rcc.ahb1enr().modify(|_, w| w.dma1en().set_bit());
+
     let mut syscfg = dp.SYSCFG.constrain(&mut rcc);
 
     println!("Led Init");
@@ -80,7 +93,7 @@ fn main() -> ! {
 
     println!("Start Loop");
     loop {
-        //wfi();
+        cortex_m::asm::wfi();
         println!("Check");
 
         if G_LED_ON.load(Ordering::Relaxed) {
--- a/examples/button.rs
+++ b/examples/button.rs
@@ -53,7 +53,7 @@ fn main() -> ! {
 
     let mut dp = stm32::Peripherals::take().expect("cannot take peripherals");
     let mut rcc = dp.RCC.constrain();
-    let mut syscfg = dp.SYSCFG.constrain();
+    let mut syscfg = dp.SYSCFG.constrain(&mut rcc);
 
     println!("Led Init");
     // Configure PA5 pin to blink LED
@@ -80,7 +80,7 @@ fn main() -> ! {
 
     println!("Start Loop");
     loop {
-        wfi();
+        //wfi();
         println!("Check");
 
         if G_LED_ON.load(Ordering::Relaxed) {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by burrbull** [comment]
What probe software do you use? `probe-rs` should do it automatically on connect.

File: examples/button.rs [hunk 1]
```diff
@@ -52,8 +51,22 @@ fn main() -> ! {
...
     let mut dp = stm32::Peripherals::take().expect("cannot take peripherals");
+
+    // Workaround for RTT when using wfi instruction
+    // Enable the debug sleep bits in DBGMCU,
+    // then enable DMA peripheral clock in AHB1ENR
+    dp.DBGMCU.cr().modify(|_, w| {
+        w.dbg_sleep().set_bit();
+        w.dbg_stop().set_bit();
+        w.dbg_standby().set_bit()
+    });
```
burrbull, line 62: `probe-rs` should do it automatically on connect.

boondocklabs:
> What probe software do you use? `probe-rs` should do it automatically on connect.

I use `probe-rs`, but it does not work with `wfi()` as is. Setting flags in `DBGMCU` explicitly is probably not necessary as I assume probe-rs does that, but enabling DMA clock on AHB is the trick to get it working. Without it, RTT does not connect.

DMA clock trick works on G431 and G474

The isb after wfi trick didn't work either. Lots of discussion about it on https://github.com/probe-rs/probe-rs/issues/350#issuecomment-1381170951

src/syscfg.rs
```diff
@@ -1,25 +1,17 @@
...
-
-            // Enable clock.
-            bb::set(&rcc.apb2enr(), 0);
-
-            // Stall the pipeline to work around erratum 2.1.13 (DM00037591)
-            cortex_m::asm::dsb();
-        }
+    fn constrain(self, rcc: &mut Rcc) -> SysCfg {
+        // Enable SYSCFG peripheral clock in APB2ENR register
+        rcc.apb2enr().modify(|_, w| w.syscfgen().set_bit());
```
usbalbin, line 14: I have not tinkered with the SysCfg, but is there any reason to handle this different from for example how the [usarts are enabled](https://github.com/stm32-rs/stm32g4xx-hal/blob/c0c0805ae8f02c2c63b97d3b9cb0d6f0782ccc20/src/serial/usart.rs#L558). So something like this?

```rust
SYSCFG::enable(rcc);
SYSCFG::reset(rcc);
```

examples/button.rs
usbalbin: @boondocklabs , if the DMA enable thing is enough, then how about we do only that? (Same for #214 i presume)

boondocklabs: I've removed the DBGMCU modification, and left the DMA enable in the button example. Tested on a G431KB without explicit DGBMCU modification and it does not hardfault as long as the DMA enable workaround is there.

src/syscfg.rs
boondocklabs: I'll take a look!

boondocklabs: Pushed a change to use Enable and Reset traits rather than direct register access

examples/i2c-fmp-as5600.rs
```diff
@@ -0,0 +1,92 @@
+//! I2C Fast Mode Plus example with an AS5600 magnetic angle sensor.
+//!
+//! This example expects the AS5600 to be connected to the I2C bus on PB7 (SDA) and PA15 (SCL),
+//! and a 24MHz HSE oscillator to configure the PLL for 168MHz system clock.
+//!
+//! The I2C bus is configured with Fast Mode Plus (FMP) enabled in SysCfg, and a 1MHz I2C clock rate.
+//!
+//! ```DEFMT_LOG=debug cargo run --release --example rand --features stm32g431,defmt -- --chip STM32G431KBTx```
```
usbalbin: This is not the `rand` example :)

examples/i2c-fmp-as5600.rs
```diff
@@ -0,0 +1,92 @@
...
+    let gpioa = dp.GPIOA.split(&mut rcc);
+    let gpiob = dp.GPIOB.split(&mut rcc);
+
+    let sda = gpiob.pb7.into_alternate_open_drain();
+    let scl = gpioa.pa15.into_alternate_open_drain();
+
+    let mut syscfg = dp.SYSCFG.constrain(&mut rcc);
+
+    // Enable Fast Mode Plus for I2C1
+    syscfg.i2c_fmp_enable::<1>(true);
```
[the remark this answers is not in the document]
burrbull: It should be moved to `i2c` module, implemented on `I2c` and take `&mut SysCfg` as parameter.

src/syscfg.rs
```diff
@@ -35,3 +27,22 @@ impl Deref for SysCfg {
...
     }
 }
+
+impl SysCfg {
+    /// Enable/disable I2C fast mode plus on the I2C bus index provided as a const generic parameter.
+    ///
+    /// Pins that are configured as I2C alternate functions will be configured as fast mode plus.
+    /// The alternate function mode of the pin must be set before FMP is enabled in SysCfg.
+    ///
+    /// When FM+ mode is activated on a pin, the GPIO speed configuration (OSPEEDR) is ignored and overridden.
```
phansel: This seems like a non-user-friendly design decision. Why go to the extent of asking users to set bitrate as e.g. 1.MHz() if it's going to be ignored?

phansel:
It seems like this PR combines two separate things, one of which seems important (fixing examples/button.rs and the SYSCFG hardfault) and one (enabling I2C FMP) which seems unrelated + slightly lower priority and also isn't ready to be merged. 

If boondocklabs:main as of a6369ca27 has the SYSCFG hardfault resolved, it should just be merged as of that commit. The I2C/FMP API change can go in a separate PR.

I can confirm this branch (as of a6369ca27) prevents the SYSCFG segfault on STM32G474RE.

PR #223, opened 2025-07-29
SPI hotfix

### LINKED ISSUE #222: spi.rs needs work
https://github.com/stm32-rs/stm32g4xx-hal/blob/main/src/spi.rs

- embedded_hal_old has not been fully removed / replaced
- hal_02 is still used in [the spi example](https://github.com/stm32-rs/stm32g4xx-hal/blob/main/examples/spi-example.rs#L58)
- little to no explanatory comments (compare: [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/spi.rs))
  - unclear what functions like [set_bidi()](https://github.com/stm32-rs/stm32g4xx-hal/blob/ee7c0103f0eb3c8617478646826e785435692c83/src/spi.rs#L204) actually do; do they convert it to full-duplex, or to half-duplex on a 3-wire interface?
- same issue: [many instances](https://github.com/stm32-rs/stm32g4xx-hal/blob/ee7c0103f0eb3c8617478646826e785435692c83/src/spi.rs#L317) of names abbreviated beyond being comprehensible

So far, I haven't been able to correctly read+write to an SPI device from stm32g474 with this HAL yet. 
- Responses (on MISO) to writes (on MOSI) are either not captured by the device or are captured incorrectly. 
- Reads, when they don't crash, make little temporal sense
  - 0x22 on the bus -> reports as [u8; 3] as [0x0c, 0x0c, 0x0c]. 
- With certain orders of operation, the microcontroller spontaneously panics.
- transfer() does not work at all. SCK sits at idle the entire time.

The intuitive API (spi.read(), spi.write()) type system is also broken in a way I have never seen before. rustc gets confused by spi.read() and isn't sure if it's a trait of the ADC, I2C, or different types of SPI.
`spi.read(&mut out_buf)` has to be fully specified as `stm32g4xx_hal::prelude::SpiBus::read` to even compile, and even then, the SPI read behavior is unreliable.
```
error[E0034]: multiple applicable items in scope
   --> src/main.rs:147:9
    |
147 |     spi.read(&mut read_buf);
    |         ^^^^ multiple `read` found
    |
note: candidate #1 is defined in the trait `_embedded_hal_adc_OneShot`
   --> /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embedded-hal-0.2.7/src/adc.rs:97:5
    |
97  |     fn read(&mut self, pin: &mut Pin) -> nb::Result<Word, Self::Error>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: candidate #2 is defined in the trait `stm32g4xx_hal::prelude::I2c`
   --> /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embedded-hal-1.0.0/src/i2c.rs:333:5
    |
333 |     fn read(&mut self, address: A, read: &mut [u8]) -> Result<(), Self::Error> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: candidate #3 is defined in the trait `stm32g4xx_hal::prelude::SpiBus`
   --> /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embedded-hal-1.0.0/src/spi.rs:446:5
    |
446 |     fn read(&mut self, words: &mut [Word]) -> Result<(), Self::Error>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: candidate #4 is defined in an impl of the trait `_embedded_hal_spi_FullDuplex` for the type `Spi<SPI, PINS>`
help: disambiguate the method for candidate #1
    |
147 -     spi.read(&mut read_buf);
147 +     _embedded_hal_adc_OneShot::read(&mut spi, &mut read_buf);
    |
help: disambiguate the method for candidate #2
    |
147 -     spi.read(&mut read_buf);
147 +     stm32g4xx_hal::prelude::I2c::read(&mut spi, &mut read_buf);
    |
help: disambiguate the method for candidate #3
    |
147 -     spi.read(&mut read_buf);
147 +     stm32g4xx_hal::prelude::SpiBus::read(&mut spi, &mut read_buf);
    |
help: disambiguate the method for candidate #4
    |
147 -     spi.read(&mut read_buf);
147 +     _embedded_hal_spi_FullDuplex::read(&mut spi, &mut read_buf);
```

techmccat:
Fixed visibility of methods set_bidi and set_tx_only, which were not supposed to be public
Fixes #222: SPI reads never happening at all because of a broken pointer cast

### DIFF
--- a/examples/spi-example.rs
+++ b/examples/spi-example.rs
@@ -18,8 +18,8 @@ use hal::{
 };
 
 use cortex_m_rt::entry;
-use log::info;
 use stm32g4xx_hal as hal;
+use utils::logger::info;
 
 #[macro_use]
 mod utils;
@@ -59,7 +59,7 @@ fn main() -> ! {
         cs.set_low();
         // transmit and receive at the same time
         spi.transfer(&mut received_msg, message).unwrap();
-        info!("{received_msg:?}");
+        info!("{:?}", &received_msg);
         cs.set_high();
 
         delay_tim2.delay_ms(1000);


examples/spi-hal-one.rs
```diff
@@ -0,0 +1,97 @@
+// This example is to test the SPI without any external devices,
+// it writes to the MOSI line and logs whatever is received on the MISO line
+//
+// MISO and MOSI should be connected to make the assertions work
```
usbalbin, line 4: From what I can tell, PA6 and PA7 are ARD_D12 and ARD_D11 on the Nucleo G474 which are next to each other. What do about adding this as a test in https://github.com/stm32-rs/stm32g4xx-hal/blob/main/tests/nucleo-g474_w_jumpers.rs ?

techmccat: I've done that and reorganized the tests a bit, I won't have access to some hardware/tools for a while so could you run the test for me?

usbalbin: Sure

usbalbin (approved):
Awesome! Thanks!

PR #224, opened 2025-08-08
Add support for true random number generator (RNG) peripheral

boondocklabs:
- Adds `RngExt` trait to constrain and enable RNG peripheral with HSI48 clock
- Implements state machine with `Rng<Stopped>` and `Rng<Running>` types
- Provides blocking and non-blocking read methods for 32-bit random values
- Implements `TryRngCore` from `rand` for interoperability with rand traits
- Includes error handling for seed and clock error conditions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ stm32-usbd = { version = "0.7.0", optional = true }
 fixed = { version = "1.28.0", optional = true }
 embedded-io = "0.6"
 stm32-hrtim = { version = "0.1.0", optional = true }
+rand = { version = "0.9", default-features = false }
 
 [dependencies.cortex-m]
 version = "0.7.7"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -20,6 +20,7 @@ stm32-usbd = { version = "0.7.0", optional = true }
 fixed = { version = "1.28.0", optional = true }
 embedded-io = "0.6"
 stm32-hrtim = { version = "0.1.0", optional = true }
+rand = { version = "0.9", default-features = false }
```
usbalbin, line 23: Should we use rand_core instead of rand?

usbalbin: Great! Thanks

usbalbin (approved):
Thank you, looks good to me!

PR #225, opened 2025-08-13
clippy

usbalbin:


boondocklabs:
Looks like we were both working on the same thing: #226 :)

PR #227, opened 2025-11-13
Expose extra USART config options and fix word length bug

Keybinder:
1. Exposes tx_invert() and rx_invert() USART/serial configuration options, which correspond to setting the TXINV and RXINV bits in USART_CR2 respectively.
2. Exposes a wordlength_7() configuration option, which corresponds to having a word length of 7.
3. Fixes a mistake in the peripheral initialisation where the M[1:0] bits were set incorrectly for word lengths of 7 and 9.


Keybinder:
Is this ok now?

PR #229, opened 2025-12-08
unify Instance traits

burrbull:
[commit 1](https://github.com/stm32-rs/stm32g4xx-hal/pull/229/commits/df2c873f825880c7ebecb15c3f7024c505b46451):
* split `GetBusFreq` on `BusClock` and `BusTimerClock`
* improve  `rcc::Instance` and use it where possible

[commit 2](https://github.com/stm32-rs/stm32g4xx-hal/pull/229/commits/7575df1236ff1ff887479a228a6de88c32807a1a) (optional):
remove `BusClock` autoimplementation. I've found that sometimes it can cause implementation conflicts

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,14 +117,41 @@ impl<RB, const A: usize> Sealed for Periph<RB, A> {}
 pub trait Ptr: Sealed {
     /// RegisterBlock structure
     type RB;
+    /// Pointer to the register block
+    const PTR: *const Self::RB;
     /// Return the pointer to the register block
-    fn ptr() -> *const Self::RB;
+    #[inline(always)]
+    fn ptr() -> *const Self::RB {
+        Self::PTR
+    }
 }
 
 impl<RB, const A: usize> Ptr for Periph<RB, A> {
     type RB = RB;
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -117,14 +117,41 @@ impl<RB, const A: usize> Sealed for Periph<RB, A> {}
...
 }
 
 impl<RB, const A: usize> Ptr for Periph<RB, A> {
     type RB = RB;
-    fn ptr() -> *const Self::RB {
-        Self::ptr()
+    const PTR: *const Self::RB = Self::PTR;
+}
+
+pub trait Steal: Sealed {
```
usbalbin, line 134: What is the reason for having lots of small traits like `Ptr`, `Steal` etc vs having one or fewer traits?

burrbull: Good point.
I could merge `Ptr` and `Steal` as both implemented for all peripherals but how to name this trait?

usbalbin: It's not a big problem, I was just curious

burrbull:
@usbalbin What is your verdict?

usbalbin (approved):
LGTM, thanks!

Issue #236, opened 2026-03-03
FlashWriter::write fails with ProgrammingError


EvanLF6768:
`FlashWriter::write` fails and returns `ProgrammingError` due to incorrect sequencing of operations in its implementation.

Page 109 of `RM0440` states that the correct sequences for a standard programming operation is:
<img width="1039" height="598" alt="Image" src="https://github.com/user-attachments/assets/de6c0575-6824-4481-b9fe-d040cf959649" />

However the current implementation does not do this (step annotations added by me).
```rs
// Step 3
// Set Page Programming to 1
self.flash.cr.cr().modify(|_, w| w.pg().set_bit());

// Step 1
while self.flash.sr.sr().read().bsy().bit_is_set() {}

// Step 4
// NOTE(unsafe) Write to FLASH area with no side effects
unsafe { core::ptr::write_volatile(write_address1, word1) };
unsafe { core::ptr::write_volatile(write_address2, word2) };

// Step 5
// Wait for write
while self.flash.sr.sr().read().bsy().bit_is_set() {}

// Step 7
// Set Page Programming to 0
self.flash.cr.cr().modify(|_, w| w.pg().clear_bit());
```

Modifying the code to follow the datasheet sequence fixes this issue (tested on an STM32G474RB):
```rs
while self.flash.sr.sr().read().bsy().bit_is_set() {}
// Set Page Programming to 1
self.flash.cr.cr().modify(|_, w| w.pg().set_bit());

// NOTE(unsafe) Write to FLASH area with no side effects
unsafe { core::ptr::write_volatile(write_address1, word1) };
unsafe { core::ptr::write_volatile(write_address2, word2) };

// Wait for write
while self.flash.sr.sr().read().bsy().bit_is_set() {}

self.flash.sr.sr().modify(|_, w| w.eop().clear_bit());
// Set Page Programming to 0
self.flash.cr.cr().modify(|_, w| w.pg().clear_bit());
```

usbalbin:
Thanks for finding and fixing this!

PR #237, opened 2026-03-03
Corrected FlashWriter::write to follow datasheet

EvanLF6768:
Slighly modify the sequence of operations in `FlashWriter::write` so that it successfully writes to internal flash (refer to issue #236 for more details). The fix has been tested on hardware.


usbalbin (approved):

usbalbin:
Awesome!

Thank you for your contribution!

